PR #41, opened 2018-07-16
[RFC] 0003 Proxy capabilities of ui5-server

RandomByte:
Add proxy capabilities to the ui5-server module.

**Read: [RFC 0003 Proxy capabilities of ui5-server](https://github.com/SAP/ui5-tooling/blob/rfc-proxy/rfcs/0003-proxy.md)**

JIRA: CPOUI5FOUNDATION-92

### DIFF
--- a/rfcs/0003-proxy.md
+++ b/rfcs/0003-proxy.md
@@ -44,9 +44,9 @@ Such resources can be an OData service endpoint but also runtime artifacts like 
 - Proxy should probably work against a single remote system
 - Useful for
     - application and library development
-        + No need to maintain/update framework dependencies locally
+        + e.g. no need to maintain/update framework dependencies locally, as they are served from the remote system
     - framework development
-        + e.g. test applications deployed in a system against a different version of OpenUI5/SAPUI5
+        + e.g. test applications deployed in a system against a different version of OpenUI5/SAPUI5 served from the local machine
 
 ### Rewriting request URL
 In all three use cases it might be required to rewrite some URLs. Examples are:
--- a/rfcs/0003-proxy.md
+++ b/rfcs/0003-proxy.md
@@ -67,6 +67,11 @@ The proxy shall be able to forward the following authentication methods:
 
 In addition, authentication for a transparent proxy (e.g. some corporate proxies) sitting between the local machine and the remote system shall be supported.
 
+### Additional requirements
+- Full TLS support (with opt-out "insecure" option)
+- Possibly WebSocket support
+- Possibly HTTP/2 support
+
 ### Configuration
 #### Project setup
 It is to be decided whether a proxy is to be configured within a single project or as a separate project.
@@ -101,6 +106,14 @@ metadata:
 proxy:
     remoteRoot: "https://..."
 ```
+
+#### General configuration options
+- Remote systems root URL
+- Optional: Paths to certificate authority files for trusted connections with systems serving self-signed/corporate certificates.
+- Optional: Host and port of transparent proxy (could also be read from environment variables, see https://github.com/SAP/ui5-server/pull/38)
+- For use case **C**:
+    + Root path for UI5 resources
+    + Root path of target application (e.g. CP deployment, Fiori Launchpad)
 
 #### Defining what requests should be processed by the proxy
 TBD
--- a/rfcs/0003-proxy.md
+++ b/rfcs/0003-proxy.md
@@ -1,0 +1,120 @@
+- Start Date: 2018-07-16
+- RFC PR: [#41](https://github.com/SAP/ui5-tooling/pull/41)
+- Issue: https://github.com/SAP/ui5-server/issues/13
+- Affected components
+    + [ ] [ui5-builder](https://github.com/SAP/ui5-builder)
+    + [x] [ui5-server](https://github.com/SAP/ui5-server)
+    + [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
+    + [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
+    + [x] [ui5-project](https://github.com/SAP/ui5-project)
+    + [ ] [ui5-logger](https://github.com/SAP/ui5-logger)
+
+# RFC 0003 Proxy capabilities of ui5-server
+## Summary
+Add proxy capabilities to the ui5-server module. It should allow to simulate an applications runtime on a remote system locally.
+
+## Motivation
+As an application or library developer, I want to be able to develop applications locally and test them in an environment as close to the production environment as possible. Meaning that resources that are not available locally are automatically retrieved from a given remote system.
+
+Such resources can be an OData service endpoint but also runtime artifacts like UI5 libraries.
+
+**Example:** I want to be able to test an application in a Fiori Launchpad environment without having anything but the application source code available locally. All other resources shall be proxied from, e.g. a NetWeaver ABAP system.
+
+## Detailed design
+### Possible use cases
+#### A: Use local resources but proxy specific requests to remote system
+- Everything is served from local resources expect requests matching a predefined list, those are proxied to a remote system
+- Possibly multiple proxies to different systems (if required)
+- Useful for
+    + OData scenarios
+        * Only requests for an OData service endpoint shall be proxied to a given system
+- Possible implementation in [PR #38](https://github.com/SAP/ui5-server/pull/38)
+
+#### B: Use remote system environment but serve specific requests from local resources
+- Everything is proxied from a remote system expect requests matching a predefined list, those are served form local resources
+- Proxy should probably work against a single remote system
+- Useful for
+    + all development
+        * Testing changes of a limited set of resources in any remote system environment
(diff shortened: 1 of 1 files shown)


RandomByte:
**Question:** When looking at the [Possible use cases](https://github.com/SAP/ui5-tooling/blob/rfc-proxy/rfcs/0003-proxy.md#possible-use-cases), is use case **A** (only proxying single paths) something widely wished and actually preferred over something like use case **C** (proxying everything that can't be found locally)?

MichaelSp:
My use-case is pretty much covered by *A* (only proxy single path)
I haven't seen *B* or *C* in the wild, so I would avoid (IMHO) any additional complexity if not necessary.
I also don't need URL-Rewriting, although I can imagine that it may come in handy later as projects grow.
For the other features:
 * Proxy auth not required
 * Basic Authentication would be nice 
 * TLS is required (insecure option would be nice)
 * Websocket is required
 * HTTP/2 would be nice  
 * Prefer "single project"-setup over "separate project"-setup

That would fit my use-case

MichaelSp:
any progress/decision yet?

matz3:
Unfortunately not

mlenkeit:
I agree with Michael: for application (and lib) development, I've only seen option A. For explicit URL rewriting, I haven't seen any need yet.

Regarding other features:
- Basic auth with credentials in either plain text (for dev system), from env vars, or from some kind of secure vault (we haven't found a good solution for that yet)
- SAML2 auth to authenticate against e.g. SCP
- SSO certificate support (we can provide details here)

mcanalesmayo:
My use case would be covered by C. Please, keep in mind that we should keep our local environments as similar as possible to the remote one (for both OData calls and UI5 resources). Take into account that there can be a lot of developers working on apps, extensions or libraries for the same remote system. It would be great if developers don't have to maintain a local copy of the framework. Same thing with the framework copy used by the CI/CD environment.

Authentication capabilities would be also nice to have, specially SAML2 for SSO.

srenwickblum:
Has this feature been abandoned?

RandomByte:
Definitely not! It's on our roadmap: https://github.com/SAP/ui5-tooling/issues/101

hschaefer123:
Hi,
we are currently using @uniorg/localneo developed from Mark (The Masch).

We use it as a replacement for the cloud web ide and therefore we need SCP support to be compatible.
It is reading the neo-app.json configuration for the paths and the missing destinations are mapped
with a small file like this

{
  "server": {
    "port": "12345",
    "hostname": "localhost",
    "open": true
  },
  "service": {
    "sapui5": {
      "useSAPUI5": true,
      "version": "1.48.5"
    }
  },
  "applications": { 
    "mylib": {
      "path": "../mylib"
    },
    "images": {
       "path": "/var/www/images"
    }
  },
 
  "destinations": {
    "SAP_BACKEND": {
      "url": "https://our_secret_sap_system.com" 
    }
  }
}

Currently it support BasicAuth on destinations ( we do not found an OData Client in NPM to support SAML with SCP).

It even handles the resources passthrouwed to CDN.

Something like this would be very handy to switch over to the new tooling.

We need Option A, because it should work similar to neo.
I think our application proxy it replaced by npm link or somethin similar and you server component,
so everything for the odata/rest destinations would be valueable.

Regards Holger

kovboyjder:
Hi, 

Just going through and thinking how we can provide the neoapp files. As the RFC states we need to be able to proxy different requests. But we need to do it in such a way we don't have to change the manifest files. This is where  the neoapp files comes in. However because we would need to pass in credentials and potentially change default clients, this is why i thought the neo-dest.json file made sense. It follows a similar pattern to WebIDE and SCP.

We could put the file paths into the yml file as neoapp and neodest. This way we could make it slightly more flexible and have it served from the tree in the index.

Would this suffice?

hschaefer123:
Hi,
this sounds similar to our needs.

Our **neo-dest.json** file looks like this

```json
{
    "server": {
        "port": "8080",
        "path": "/webapp/index.html",
        "open": true
    },
    "service": {
        "sapui5": {
            "useSAPUI5": true,
            "version": "1.56.5"
        }
    },
    "destinations": {
        "ODATA_SRV": {
            "url": "https://my.domain.de/odata_srv",
            "auth": "user@domain.de:12345678"
        }
    }
}

```

but you maybe need only a portion of this.

The service section is for easy customization off the always used /resources and /test-resources services inside neo-app.json. We auto link to CDN while using SAPUI5 by default or OPENUI5 if useSAPUI5 is false. By default returning the latest UI5 version or an explicit one (the same way like this can be configured in the Web IDE on project level).

I think currently the ui5/tooling does not support preload libs and always fetches all source classes, but hopefully this will also be an option in the future to load libs! 

Using it this way, we are currently completely web-ide/neo compatible. 

Regards Holger

kovboyjder:
I've followed Petr Plenkov's example in my edition of the neo-app.json and neo-dest.json. However these would need to be added to the yaml file for proper config instead of being hardcoded in the js file. Just want to make sure it makes sense before i try to do it and these changes would fulfill the RFC.

gregorwolf:
I'm still wondering why this issue isn't solved using the existing @sap/approuter module? @vobu just described in [cf for your pocket](https://blogs.sap.com/2019/07/04/cf-for-your-pocket/) how to use it locally.

RandomByte:
> I'm still wondering why this issue isn't solved using the existing @sap/approuter module? @vobu just described in [cf for your pocket](https://blogs.sap.com/2019/07/04/cf-for-your-pocket/) how to use it locally.

Main issue for us is that it is not open source. Also it does not easily integrate into existing server solutions like ours. It's rather the other way around, as the approuter project provides extension hooks itself. We'd therefore rather leave an integration of the UI5 Tooling *into* the approuter up to the community for now.

mlenkeit:
@RandomByte @matz3 is there any update on this? After all, this topic has been in discussion for over a year now, although it's probably the most important feature that's missing in the new UI5 tooling when it comes to adopting it for app development.

Maybe it's worth considering a change in strategy from *building a handle-all-cases solution that's convenient to use* to *building the minimum functionality (MVP) for the 80% case*. You can still add sugar coating, such as getting the information from `neo-app.json` or `xs-app.json`, taking SCP destinations into account, etc. at a later point in time.

RandomByte:
fyi, we have published a possible implementation of a somewhat NetWeaver ABAP / Gateway focused "proxy mode" as part of the standard UI5 Tooling: https://github.com/SAP/ui5-server/pull/223

This is a PoC that was initially developed quite some time ago and now finally got published. However, I think its scope is limited to a certain set of UI5 developers working with NetWeaver ABAP systems

preetamkajalrout:
If anyone is still looking for a proxy plugin that supports neo-app.json. You can give it a try to plugin I created for my use-case:
[ui5-middleware-destination](https://www.npmjs.com/package/ui5-middleware-destination)

Feedback are welcome 👍

RandomByte:
It's been a while and I would like to do a quick pulse check:

**Does anybody here still have proxy-related requirements that are not, or can not be covered by [custom server middleware](https://sap.github.io/ui5-tooling/pages/extensibility/CustomServerMiddleware/)?**

Today, there are multiple proxy extensions available in the [ui5-community organization](https://ui5-community.github.io/ui5-ecosystem-showcase/#available-middlewares). Others can be found on GitHub and npm, like the one @preetamkajalrout [mentioned](https://github.com/SAP/ui5-tooling/pull/41#issuecomment-628102475).

---

In UI5 Tooling we always struggled with the different requirements for different proxy scenarios. Instead of implementing a "standard" proxy functionality, we are now rather thinking of enhancing the extensibility mechanisms to offer for example helper-functions to custom proxy implementations. But also here, we are not sure what is missing or required. We therefore highly appreciate your feedback.

LukasHeimann:
Hi @RandomByte,

I'm trying to build a proxy that can cover websockets as well. This may be just lacking in the documentation, but I have a feeling like this isn't possible with middlewares today.

My middleware looks like this:

```js
const url = require('url');
const httpProxy = require('http-proxy');

module.exports = function({resources, options}) {
    const proxy = httpProxy.createProxyServer({
        changeOrigin: true, // change the origin to the target host
        ignorePath: true, // we combine the full target ourselves
        secure: false // don't try to valide (internal) certificates
    });

    proxy.on('error', (err, req, res) => {
        console.error(err);
        res.end(err.message);
    });

    return function backendProxy(req, res, next) {
        const proxyTarget = url.resolve(options.configuration.url, req.path);
        proxy.web(req, res, { target: proxyTarget });
        return;
    }
};
```

This doesn't cover websockets. As per [http-proxy's documentation](https://www.npmjs.com/package/http-proxy#proxying-websockets), you'd need to attach to the httpserver to do that, i.e.:

```js
ui5InternalHttpServer.on('upgrade', function (req, socket, head) {
  // TODO: Check if the middleware's `mountPath` matches -- unknown where to read that from
  proxy.ws(req, socket, head);
});
```

If there is a way to do this right now with middlewares, I'd like to hear about it, but it seems to not fit their design, tbh.

Thanks for helping me out here!
Kind regards
Lukas

PR #114, opened 2019-01-07
[INTERNAL] package.json: Add workaround for http downgrade of package-lock.json URLs

RandomByte:
See https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/44 and https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/16


RandomByte:
Nice 🎉

PR #135, opened 2019-02-15
[RFC] Add RFC template file

RandomByte:


rfcs/0000-template.md
```diff
@@ -0,0 +1,44 @@
+- Start Date: YYYY-MM-DD
+- RFC PR: -
+- Issue: -
+- Affected components
+    + [X] [ui5-builder](https://github.com/SAP/ui5-builder)
```
matz3: Maybe don't "check" any component by default. If someone forgets to check them it's easier to see when no component is checked.

RandomByte: I'm not sure whether people will otherwise know that an "X" needs to be used to "check" a component...

RandomByte:
Still missing: Updated CONTRIBUTING.md

rfcs/0000-template.md
matz3: You can also check them within the markdown preview just like normal checkboxes.

RandomByte: Not in your local IDE

matz3: Hm okay. What about an HTML comment that describes what to do? It will be visible but is not rendered in the markdown. This way we might add other disclaimers which should not be part of the final document.

RandomByte: Great idea 👍

matz3:
@RandomByte what exactly did you want to put into the CONTRIBUTING.md?
Regardless of that I would be ok with merging the template right now.

RandomByte:
A reference to it. But yeah, fair enough. With your approval I would merge 👍

RandomByte:
I already had a pending commit locally. Should be complete now 👍

rfcs/0000-template.md
RandomByte: Done

PR #151, opened 2019-04-25
[RFC] 0005 Server Middleware Extensibility

svbender:
Read the current draft here: **[0005-server-middleware-extensibility.md](https://github.com/SAP/ui5-tooling/blob/rfc-server-middleware-extensibility/rfcs/0005-server-middleware-extensibility.md)**

### DIFF
--- a/rfcs/0005-server-middleware-extensibility.md
+++ b/rfcs/0005-server-middleware-extensibility.md
@@ -11,14 +11,14 @@
 
 # RFC 0005 Server Middleware Extensibility
 ## Summary
-Add a feature to load custom or third party server middleware from `npm` dependencies. 
+Add a feature to load custom server middleware from a projects dependencies or configuration. 
 
 ## Motivation
 Currently the UI5 Server comes with a fixed set of middleware, which can not be extended.
 A developer may want to handle requests differently. For example the developer wants to add various headers to a response
 or parse data of a POST request in a specific way.
 
-Custom or existing third party middleware can provide the required behavior.
+Custom middleware can provide the required behavior.
 
 ## Detailed design
 ### Configuration
@@ -28,9 +28,6 @@ server started in this project. Custom middleware configuration of dependencies 
 
 A middleware may be executed before or after any other middleware.
 This shall be configurable in a simple but less generic way.
-
-Custom and third party server middleware modules (from `npm` dependencies) are treated differently and have their own
-configuration sections (`customMiddleware` and `npmMiddleware`).
 
 A project configuration might look like this:
 ```yaml
@@ -45,19 +42,14 @@ server:
       afterMiddleware: compression
       configuration:
         debug: true
-  npmMiddleware:
-    - name: helmet
-      mountPath: /
-      beforeMiddleware: cors
 ```
 
-In the above sample the middleware `cors` and `compression` are already included by the UI5 Server.
+In the above sample the middleware `compression` is already included by the UI5 Server.
 
-When serving the application `my.application`, this will execute the third party middleware `helmet` before
-`cors` and the custom middleware `myCustomMiddleware` after `compression`.
+When serving the application `my.application`, this will use the custom middleware `myCustomMiddleware` after `compression`.
 
 For custom server middleware, there can be optional `configuration` parameters which are passed
-directly to it with `options.configuration`. This is not possible for third party server middleware. 
+directly to it with `options.configuration`.
 
 An optional `mountPath` for which the middleware function is invoked, can be provided for both sections.
 It will be passed to the `app.use` call (see [express API reference](https://expressjs.com/en/4x/api.html#app.use)).
@@ -74,13 +66,6 @@ like "extensions", an additional attribute "kind" is added to the ui5.yaml.
 
 A custom middleware will consist of at least a ui5.yaml defining it as an extension and
 a JavaScript implementation.
-
-A third party middleware which do not contain a ui5.yaml will be taken as it is.
-In this case the server expects the [default express API](https://expressjs.com/en/guide/writing-middleware.html). 
-The configured module name will be required directly by the UI5 Server, i.e. it will not be treated as a
-`server-middleware` extension.  
-
-If the third party middleware requires additional parameters, a custom middleware wrapper needs to be implemented.
 
 #### Example middleware extension
 **`ui5.yaml`**:
@@ -142,10 +127,7 @@ The UI5 Server will detect the custom middleware configuration of the project my
 into the servers start process to be available for the following HTTP requests.
 
 ##### Execution order
-The order of the middleware is important. In the following it is decribed in which order the middleware is executed.
-1. The section `customMiddleware` before `npmMiddleware`. This means, if a custom and a third party middleware from
-should be both applied before or after the same middleware, the custom middlware will be executed first.  
-2. Multiple middleware configurations within the same section are executed in the provided order. 
+The order of the middleware is important. Multiple middleware configurations are applied in the provided order. 
 
 ### Server middleware implementation
 A custom middleware implementation needs to return a function with the following signature (written in JSDoc):
--- a/rfcs/0005-server-middleware-extensibility.md
+++ b/rfcs/0005-server-middleware-extensibility.md
@@ -129,7 +129,7 @@ server:
 ---
 specVersion: "1.0"
 kind: extension
-type: middleware
+type: server-middleware
 metadata:
   name: myCustomMiddleware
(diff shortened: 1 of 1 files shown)


rfcs/0005-server-middleware-extensibility.md
```diff
@@ -0,0 +1,154 @@
...
+UI5 Build and Development Tooling. Examples for other extensions include "Shims" ([RFC 0002](https://github.com/SAP/ui5-tooling/blob/master/rfcs/0002-project-shims.md)), tasks 
+([RFC 0004](https://github.com/SAP/ui5-tooling/blob/master/rfcs/0004-simple-build-extensibility.md)), and translators.
+
+To separate "UI5 Projects" (i.e. things that represent UI5-artifacts for the browser) from tooling specific things
+like "extensions", an additional attribute "kind" is added to the ui5.yaml.
+
+A custom middleware will consist of at least a ui5.yaml defining it as an extension and
+a JavaScript implementation.
+
+A third party middleware which do not contain a ui5.yaml will be taken as it is. In this case the server expects
```
RandomByte: Not sure how to understand this. Further description (maybe an example?) of this scenario is needed

rfcs/0005-server-middleware-extensibility.md
```diff
@@ -0,0 +1,154 @@
...
+the [default express API](https://expressjs.com/en/guide/writing-middleware.html). 
+
+#### Example middleware extension
+**`ui5.yaml`**:
+```yaml
+specVersion: "0.1"
+kind: extension
+type: middleware
```
RandomByte: Not sure whether `middleware` or `server-middleware` is better 🤔

Middlewares *could* also exist in the builder. I mean they don't, but this still might confuse the end user.

rfcs/0005-server-middleware-extensibility.md
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * Custom ui5-server middleware example
+ *
+ * @param {Object} resourceCollections Contains the resource reader or collection to access project related files
+ * @param {AbstractReader} resourceCollections.source Reader to access the projects source files.
+ * @param {AbstractReader} resourceCollections.dependencies Reader or collecation to access the projects dependencies.
+ * @param {ReaderCollectionPrioritized} resourceCollections.combo Contains a prioritized collection to read and write all project files. 
+ * @returns {Function} Returns a middleware closure.
+ */
+module.exports = function({resourceCollections, tree}) {
```
RandomByte: The configuration that can be supplied in the ui5.yaml is not passed to the middleware here.

In general I would prefer to keep the signature aligned with that of the custom tasks: https://github.com/SAP/ui5-project/blob/master/docs/BuildExtensibility.md#task-implementation

The project tree is currently missing there but is also not required for most server middlewares. We can discuss this in person later.

rfcs/0005-server-middleware-extensibility.md
```diff
@@ -0,0 +1,154 @@
+- Start Date: 2019-04-23
+- RFC PR: -
+- Issue: -
```
RandomByte: `https://github.com/SAP/ui5-server/issues/113` (already forgot about this, sorry 😅)

sebbi08:
hi im currently missing som topic about middlewares from dependencies like we have it in the builder. 

In the builder it is currently that every application/library can define there own build steps and they run only for the coresponding like `library-a` needs to be transpiled with bable `library-b`needs no transpiling because it was written in ES5 but the app using this two libs was written in Typescript and need to be transpiled from this.

So when we have to some how also respect this for the middlewares because we alway develop on the source of the libraries and so the corresponding middlewares need to run only on requests to there own source.

svbender:
@sebbi08 Thanks for your comment.

It is currently not planned to detect/include server middleware of dependent projects.

However, the custom server middleware can be located in a dependency. Assuming the extension is defined in that projects ui5.yaml file.

This dependency can then be included in your project.
To use the custom server middleware, you only need to configure it in your projects ui5.yaml (where the UI5 Server is started).

Regards,
Sven

rfcs/0005-server-middleware-extensibility.md
matz3: What would be a reason for returning a promise for middleware functions?
Do you just want to align the signature with the tasks or do you see use cases where this might be required?
I would rather stick with our existing middleware signature as we would otherwise need to change our own middlewares in an incompatible way, or have them in a different signature which is confusing IMO.

You can still do async stuff in your middleware before accepting the first request.

sebbi08:
@svbender thats true but lets asume we dont have a "simple" middleware like transpiling in `library-a` but rather some complex stuff that need acces to files of `library-a`. with the syntax

```js
module.exports = async function({workspace, dependencies, options}) {
    return function (req, res, next) {
      // middleware code...
      next();
    }
};
```

while developing the files need would be found in the `workspace` collection and when used in the project they have to be searched in the `dependencies`

Thodd:
@sebbi08: We reconsidered this feature and tried to work in your feedback.
Would you please take another look at the new API proposal concerning the Reader arguments?

sebbi08:
@Thodd thats a nice way. now we can write the middleware in the dependencie and also reuse it from there. thanks

cschuff:
First of all once again: Thanks for your great work!

Any updates on this RFC? Is it already on the roadmap? When can we expect some progress?

BR
Chris

RandomByte:
@cschuff it's on the [roadmap](https://github.com/SAP/ui5-tooling/issues/101). Maybe @ecker can share some info on the timeline?

ecker:
@cschuff Hi Christian, LTNS 😉, middleware extensibility is one of our major topics, and while finishing some other important tooling topics, we'd like to continue on it soon, i.e. the next weeks/month(s).

cschuff:
Hi Andreas 👋, that is great news. ui5-tooling really is a major improvement to the ui5 dev ecosystem!

RandomByte:
Demo: https://github.com/SAP/openui5-sample-app/pull/50

Based on WIPs:
- https://github.com/SAP/ui5-project/pull/190
- https://github.com/SAP/ui5-server/pull/200

matz3:
I don't quite like the idea of having the both sync and async options for the middleware functions.
When using middlewares programatically, you would always need to check whether the middleware is async or not (or always wrap it with `Promise.resolve()`). Also using async middlewares is cumbersome in general as you can't easily use them in an existing scenario which needs to be executed sync. For sure you could register the middleware once the Promise resolves, but this leads to an initial timeframe where the middleware is not mounted yet but the server might be already started.

So I would like to go back to my initial opinion to make the middlewares always return a function, not a Promise.
In case a middleware needs to do some async stuff initially (which we don't have right now, or?) it might just queue the incoming requests and handle them once the initialization has finished by waiting for an initial Promise to resolve.
Yes, this makes such middlewares a bit more complex to write/maintain, but I see a bigger advantage on the consumer side.

RandomByte:
While implementing https://github.com/SAP/ui5-server/pull/200, knowing that the middleware modules *could* return a promise lead to some good design choices allowing them to do so in the future. We could definitely start with synchronous middleware modules now and wait for someone to come up with a good use case for returning a promise in the createMiddleware function.

Again, my idea was that a middleware could do some initial processing with the `resources` provided to the createMiddleware function before being ready to serve requests. But as we are still lacking a real scenario for this I'm fine restricting it to sync for now. But I would like to hear @Thodd`s opinion on this too.

RandomByte:
We decided to remove the `npmMiddleware` configuration option from the RFC for the time being.

It was meant to offer a quick way to integrate generic middleware modules that do not make use of UI5 Tooling specific APIs and therefore do not require any UI5 Server specific implementation. Also, they must not require any configuration to be passed to them. An example is the [helmet](https://www.npmjs.com/package/helmet) module.

To consume such modules with the current state of the RFC, you now need to wrap such modules into a custom middleware extension.

Please let us know if you would like to see this additional feature in a future version of the UI5 Server.

RandomByte:
Implementation is done and released with UI5 CLI [v1.6.0](https://github.com/SAP/ui5-cli/releases/tag/v1.6.0). Functionality is as described in this RFC.

Dedicated documentation is pending as part of https://github.com/SAP/ui5-tooling/pull/160 and should be published sometime tomorrow or early next week

Therefore merging this RFC 🎉 

**Thanks to everyone for your feedback!**

RandomByte:
Documentation now available at https://sap.github.io/ui5-tooling/pages/extensibility/CustomServerMiddleware/ 🐴

PR #160, opened 2019-07-03
[INTERNAL] Consolidate UI5 Tooling Documentation

RandomByte:
Use [MkDocs](https://www.mkdocs.org/) to provide a consolidated documentation for all UI5 Tooling
modules.

To test the mkdocs part locally using the provided development server:
```sh
npm run docs
```
Then open http://localhost:8000

To also test with JSDoc, everything needs to be built and served statically:
```sh
npm run docs-generate
```

Deployment will be switched from Travis to GitHub Actions


RandomByte:
CC @Thodd

PR #168, opened 2019-07-15
[RFC] 0007 Properties File Encoding

tobiasso85:
Add a fix for properties file encoding when they are part of a build or served such that they can always be interpreted as UTF-8 encoded files.

Read: [0007-properties-file-encoding.md](https://github.com/SAP/ui5-tooling/blob/rfc-properties-file-encoding/rfcs/0007-properties-file-encoding.md)

### DIFF
--- a/rfcs/0006-properties-file-encoding.md
+++ b/rfcs/0006-properties-file-encoding.md
@@ -1,0 +1,81 @@
+- Start Date: 2019-07-15
+- RFC PR: -
+- Issue: [#161](https://github.com/SAP/ui5-tooling/issues/161)
+- Affected components
+    + [x] [ui5-builder](https://github.com/SAP/ui5-builder)
+    + [x] [ui5-server](https://github.com/SAP/ui5-server)
+    + [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
+    + [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
+    + [ ] [ui5-project](https://github.com/SAP/ui5-project)
+    + [ ] [ui5-logger](https://github.com/SAP/ui5-logger)
+
+
+# RFC 0006 Properties File Encoding
+## Summary
+Properties files (`*.properties`) should be encoded in UTF-8 such that the client properly can consume it.
+
+## Motivation
+Currently the properties files are encoded in ISO-8859-1 but served as UTF-8 which leads to the problem that special characters are not displayed correctly.
+The user wants to be able to use properties files with ISO-8859-1 encoding and also have the option to specify the encoding.
+
+## Detailed design
+### Configuration
+A configuration for the task/server should be provided such that the source encoding can be specified.
+There should be 2 options:
+* `ISO-8859-1` (non ASCII characters will be escaped with the unicode sequence `\uXXXX`)
+* `UTF-8` (no character escaping will take place)
+
+The default is `ISO-8859-1` for compatibility reasons.
+
+Umlaut Example:
+
+| Umlaut   |	Unicode         |
+|----------|--------------------|
+| `Ä`, `ä` | `\u00c4`, `\u00e4` |
+| `Ö`, `ö` | `\u00d6`, `\u00f6` |
+| `Ü`, `ü` | `\u00dc`, `\u00fc` |
+| `ß` 	   | `\u00df `          |
+
+This should be configured within the .ui5.yml file.
+
+Example:
+```yaml
+specVersion: "1.0"
+type: application
+metadata:
+  name: my.application
+resources:
+  propertiesFiles:
+    - encoding: "UTF-8"
+```
+
+The resources section within the configuration can be consumed by `ui5-builder` and `ui5-server`.
+
+### Build Task
+The `ui5-builder` should offer a task called `encodePropertiesFiles` which escapes all special characters in unicode using the unicode escape sequence `\uXXXX`.
+It should use a processor called `stringEscaper` which escapes special characters in files and is used within the task to operate only on `*.properties` files.
+The processor offers a function called `escapeNonAsciiAsUnicode` which performs the unicode escaping.
+The task should be run first (before `replaceCopyright`) for all types.
+This ensures that the properties files can always be consumed.
+Each build output contains then the escaped properties files.
+This task should automatically be integrated into the build process and can be reused by the server.
+
+### Server part
+The `ui5-server` offers the capability to re-use the ui5-builder processor `stringEscaper` used by the build task to escape these characters.
+The served content can always be interpreted as UTF-8 and therefore be properly consumed.
+There should be a special treatment in [serveResources](https://github.com/SAP/ui5-server/blob/master/lib/middleware/serveResources.js#L42) middleware which uses the configuration and the processor to serve properties files such that they can be interpreted as UTF-8 encoded files.
+
+
+## How we teach this
+- Documentation about how to specify the properties files encoding option
+- Explanation of the Escaping with examples
+
+## Drawbacks
+
+
+## Alternatives
+ - Check the file contents and guess its encoding
+   - Use response content-type header attribute
+   - Only escape if required
+
+## Unresolved Questions and Bikeshedding
--- a/rfcs/0007-properties-file-encoding.md
+++ b/rfcs/0007-properties-file-encoding.md
@@ -84,18 +84,16 @@ A static function (`nonAsciiEscaper.getEncodingFromAlias`) should be added to ma
 
 ### Task
 
-A new task `escapeNonAsciiCharacters` should be added which receives the source encoding and a pattern to glob files within the workspace.
+A task `escapeNonAsciiCharacters` should be added which receives the source encoding and a pattern to glob files within the workspace.
 It uses the `nonAsciiEscaper.getEncodingFromAlias` function to map the source encoding (based on the project configuration values) to a valid encoding for the processor.
 
 ### Types
 
-The task operates on `*.properties` files using the processor.
-The task should run first (before `replaceCopyright`) for all types.
+Both `application` and `library` types should be enhanced.
 
-This ensures that the properties files can always be consumed independent of the source encoding.
-Each build output should contain the escaped properties files.
+The formatters should take the [configuration](#Configuration) into account and apply the default.
 
-The new standard task should automatically be integrated into the build process and the underlaying processor can be reused by the `ui5-server`.
+The builders should execute the `escapeNonAsciiCharacters` task as the very first task and pass the propertiesFileSourceEncoding from the project configuration. The pattern should include all properties files (`/**/*.properties`).
 
 ### Server
 
--- a/rfcs/0007-properties-file-encoding.md
+++ b/rfcs/0007-properties-file-encoding.md
@@ -39,9 +39,9 @@ But as there are existing tools and server middleware which explicitly expect th
 A configuration for the `ui5-builder` tasks and the `ui5-server` should be provided such that the source encoding of `*.properties` files can be specified.
 
 This configuration is set on project level, so that multiple projects with different encodings function idependently.
-Resources create as part of the project contain a reference to the project, which allows to read the expected source encoding for a single resource.
+Resources created as part of the project contain a reference to the project, which allows to read the expected source encoding for a single resource.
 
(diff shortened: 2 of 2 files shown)


rfcs/0006-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+resources:
+  propertiesFiles:
+    - encoding: "UTF-8"
+```
+
+The resources section within the configuration can be consumed by `ui5-builder` and `ui5-server`.
+
+### Build Task
+The `ui5-builder` should offer a task called `encodePropertiesFiles` which escapes all special characters in unicode using the unicode escape sequence `\uXXXX`.
+It should use a processor called `stringEscaper` which escapes special characters in files and is used within the task to operate only on `*.properties` files.
```
codeworrior: I find the mixture of encode/escape confusing.

rfcs/0006-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+The processor offers a function called `escapeNonAsciiAsUnicode` which performs the unicode escaping.
+The task should be run first (before `replaceCopyright`) for all types.
+This ensures that the properties files can always be consumed.
+Each build output contains then the escaped properties files.
+This task should automatically be integrated into the build process and can be reused by the server.
+
+### Server part
+The `ui5-server` offers the capability to re-use the ui5-builder processor `stringEscaper` used by the build task to escape these characters.
+The served content can always be interpreted as UTF-8 and therefore be properly consumed.
```
codeworrior: I would also mention that the pure ASCII encoding ensures that the files work on other platforms, too. No matter whether the servers there expect the *.properties files to be in ISO8859-1 or UTF-8 or something different.

rfcs/0006-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+- Documentation about how to specify the properties files encoding option
+- Explanation of the Escaping with examples
+
+## Drawbacks
+
+
+## Alternatives
+ - Check the file contents and guess its encoding
+   - Use response content-type header attribute
+   - Only escape if required
```
codeworrior: What does 'if required' exactly mean? The above escaping is also only applied for characters that need it. I guess this alternative more means 'when the real file encoding and the transport encoding don't match' or something like that.

rfcs/0007-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+Umlaut Example:
+
+| Umlaut   |	Unicode         |
+|----------|--------------------|
+| `Ä`, `ä` | `\u00c4`, `\u00e4` |
+| `Ö`, `ö` | `\u00d6`, `\u00f6` |
+| `Ü`, `ü` | `\u00dc`, `\u00fc` |
+| `ß` 	   | `\u00df `          |
+
+This should be configured within the .ui5.yml file.
```
Thodd: One dot too much :)
.  ui5.yaml

rfcs/0007-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+    - encoding: "UTF-8"
+```
+
+The resources section within the configuration can be consumed by `ui5-builder` and `ui5-server`.
+
+### Build Task
+The `ui5-builder` should offer a task called `encodePropertiesFiles` which escapes all special characters in unicode using the unicode escape sequence `\uXXXX`.
+It should use a processor called `stringEscaper` which escapes special characters in files and is used within the task to operate only on `*.properties` files.
+The processor offers a function called `escapeNonAsciiAsUnicode` which performs the unicode escaping.
+The task should be run first (before `replaceCopyright`) for all types.
```
Thodd: Hm... isn't adding new standard tasks always changing the ordering of the custom task as well?
So if somebody already wanted to be "first" and has registered a custom task before `replaceCopyright`, they are now only 2nd...

Not sure if we see that as a problem, but the ordering of tasks seems important since their results might depend on eachother, right?

rfcs/0006-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+A configuration for the task/server should be provided such that the source encoding can be specified.
+There should be 2 options:
+* `ISO-8859-1` (non ASCII characters will be escaped with the unicode sequence `\uXXXX`)
+* `UTF-8` (no character escaping will take place)
+
+The default is `ISO-8859-1` for compatibility reasons.
+
+Umlaut Example:
+
+| Umlaut   |	Unicode         |
```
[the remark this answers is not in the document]
Thodd: Done.

rfcs/0006-properties-file-encoding.md
Thodd: I added this to the `ui5-builder` section.

rfcs/0006-properties-file-encoding.md
Thodd: Done.

rfcs/0006-properties-file-encoding.md
Thodd: Done.

rfcs/0006-properties-file-encoding.md
```diff
@@ -0,0 +1,81 @@
...
+    + [x] [ui5-server](https://github.com/SAP/ui5-server)
+    + [ ] [ui5-cli](https://github.com/SAP/ui5-cli)
+    + [ ] [ui5-fs](https://github.com/SAP/ui5-fs)
+    + [ ] [ui5-project](https://github.com/SAP/ui5-project)
+    + [ ] [ui5-logger](https://github.com/SAP/ui5-logger)
+
+
+# RFC 0006 Properties File Encoding
+## Summary
+Properties files (`*.properties`) should be encoded in UTF-8 such that the client properly can consume it.
```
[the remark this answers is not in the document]
Thodd: I tried to rephrase the text to reflect the ASCII part.
I added a short reference to the SAP use-case in the Motivation section.

rfcs/0007-properties-file-encoding.md
Thodd: Done.

rfcs/0007-properties-file-encoding.md
```diff
@@ -0,0 +1,94 @@
...
+The resources section within the configuration can be consumed by `ui5-builder` and `ui5-server`.
+
+### Build Task
+
+The `ui5-builder` should offer a new standard task called `escapeNonAsciiCharacters` which escapes all special characters in unicode using the unicode escape sequence `\uXXXX`.
+It should use a processor called `stringEscaper` which escapes special characters in files and is used within the task to operate only on `*.properties` files.
+The task should be run first (before `replaceCopyright`) for all types.
+This ensures that the properties files can always be consumed.
+Each build output should contain then the escaped properties files.
+This ensures taht the now purely ASCII based `*.properties` files can be used on other platforms too.
```
RandomByte: > taht

Typo

rfcs/0007-properties-file-encoding.md
```diff
@@ -0,0 +1,94 @@
...
+## Detailed design
+
+### Configuration
+
+A configuration for the `ui5-builder` tasks and the `ui5-server` should be provided such that the source encoding can be specified.
+
+There should be 2 options:
+
+- `ISO-8859-1` (non ASCII characters will be escaped with the unicode sequence `\uXXXX`)
+- `UTF-8` (no character escaping will take place)
```
RandomByte: > `UTF-8` (no character escaping will take place)

If a server treats the resulting file as being `ISO-8859-1` encoded, we get misinterpreted characters again I suppose.

Don't we need to escape all non-ASCII characters in the UTF-8 case as well? It seems that this is also how https://github.com/SAP/ui5-builder/pull/293 is currently implemented.

RandomByte: Also see Matthias' comment [here](https://github.com/SAP/ui5-builder/pull/293/files/6504c6efced1d365e39b213ab6e5b94e40234e99#diff-02d343d30f4acdc0de962fcab14736a2)

rfcs/0007-properties-file-encoding.md
tobiasso85: done

rfcs/0007-properties-file-encoding.md
tobiasso85: done

tobiasso85:
would it make sense to later on introduce an auto-detection feature such as:
https://github.com/sonicdoe/detect-character-encoding
?

RandomByte:
> would it make sense to later on introduce an auto-detection feature such as: sonicdoe/detect-character-encoding?

I like that this module also returns a confidence score. But this already shows the problem: some character codes exist in multiple encodings.

I definitely don't want to identify an encoding wrong, which would result in the wrong characters being used in a translation file. So for me, the confidence score would need to be near 100% to be acceptable to work with.

But even in cases where a file can be identified 100% correctly, if it changes and the confidence score drops we would need to prompt the developer to specify the encoding anyways. This situation might occur in case of automated commits from a translation system that add new special characters. This would then break the build. I don't like that possibility 😕

codeworrior:
@RandomByte: this was exactly the reason why we always tried to avoid sniffing in the Java tooling and server, too.

tobiasso85:
> > would it make sense to later on introduce an auto-detection feature such as: sonicdoe/detect-character-encoding?
> 
> I like that this module also returns a confidence score. But this already shows the problem: some character codes exist in multiple encodings.
> 
> I definitely don't want to identify an encoding wrong, which would result in the wrong characters being used in a translation file. So for me, the confidence score would need to be near 100% to be acceptable to work with.
> 
> But even in cases where a file can be identified 100% correctly, if it changes and the confidence score drops we would need to prompt the developer to specify the encoding anyways. This situation might occur in case of automated commits from a translation system that add new special characters. This would then break the build. I don't like that possibility 😕

This could also lead then to a problem when some files are served with one encoding and others with another one^^

rfcs/0007-properties-file-encoding.md
```diff
@@ -13,27 +13,39 @@
...
 ## Detailed design
 
 ### Configuration
 
 A configuration for the `ui5-builder` tasks and the `ui5-server` should be provided such that the source encoding of `*.properties` files can be specified.
 
-Encodings should be supported according to the [Encoding spec](https://encoding.spec.whatwg.org/).
-For more information check out: [Buffers and Character Encodings](https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings)
+This configuration is set on project level, so that multiple projects with different encodings function idependently.
+Resources create as part of the project contain a reference to the project, which allows to read the expected source encoding for a single resource.
```
RandomByte: > create**d**

rfcs/0007-properties-file-encoding.md
```diff
@@ -13,27 +13,39 @@
...
+
+Altought there are lots of different encodings, the configuration on projects only foresees the two relevant encodings.
```
RandomByte: > the configuration on projects

**of** projects?

RandomByte: > foresees ~the~ two relevant

RandomByte:
Implementation:
- ~https://github.com/SAP/ui5-builder/pull/293~
- ~https://github.com/SAP/ui5-server/pull/214~
- https://github.com/SAP/ui5-project/pull/207

Documentation:
- https://github.com/SAP/ui5-tooling/pull/170

RandomByte:
This RFC looks pretty much done to me.

Any further feedback from any of you @Thodd, @tobiasso85, @codeworrior?

RandomByte:
**TODO:**
- [ ] Look into https://github.com/SAP/ui5-server/pull/214/files#r308795922
- [ ] Finish documentation: https://github.com/SAP/ui5-tooling/pull/170

tobiasso85:
> This RFC looks pretty much done to me.
> 
> Any further feedback from any of you @Thodd, @tobiasso85, @codeworrior?

Looks good to me

matz3:
Also LGTM

RandomByte:
Please **squash and merge!**

RandomByte:
Demo: https://github.com/SAP/openui5-sample-app/pull/54

RandomByte:
Released with UI5 CLI `v1.7.0`

PR #172, opened 2019-07-29
[INTERNAL] Consolidate UI5 Tooling Documentation part II

RandomByte:


mkdocs.yml
```diff
@@ -7,8 +7,14 @@ site_dir: 'site' # default
 nav:
     - Home: index.md
     - Getting Started: pages/GettingStarted.md
+    - Working with the UI5 Tooling:
```
matz3: That's quite a long title..

RandomByte: Yes

RandomByte: Can we live with it for the time being?

RandomByte: Renamed to "Workflows"

PR #180, opened 2019-09-02
[INTERNAL] Convert main.workflow to Actions V2 yml files

RandomByte:
As described here: https://help.github.com/en/articles/migrating-github-actions-from-hcl-syntax-to-yaml-syntax


.github/workflows/push.yml
```diff
@@ -0,0 +1,29 @@
+on: push
+name: Deploy docs
+jobs:
+  isOnMaster:
```
RandomByte: TODO: Look into all those `isOnMaster` strings

PR #217, opened 2020-03-23
Add Spec version 2.0

svbender:


docs/pages/Configuration.md
```diff
@@ -308,13 +308,14 @@ All changes are documented below.
 
 ### Compatibility Matrix
 
-Unless otherwise noted in the table below, the UI5 Tooling modules are backward compatible in the means that for example UI5 CLI v2.0 will still be able to handle a project that is using Specification Version `1.0`.
+Unless otherwise noted in the table below, the UI5 Tooling modules are backward compatible in the means that for example UI5 CLI v3.0.0 will still be able to handle a project that is using Specification Version `1.0`.
```
RandomByte: After talking to Matthias, I think we should change "UI5 CLI v3.0.0 will still" to "UI5 CLI v3.0.0 **might** still"

docs/pages/Configuration.md
```diff
@@ -330,3 +331,16 @@ Version 1.0 projects are supported by [UI5 CLI](https://github.com/SAP/ui5-cli)
 Adds support for the `theme-library` type.
 
 Version 1.1 projects are supported by [UI5 CLI](https://github.com/SAP/ui5-cli) v1.13.0 and above.
+
+### Specification Version 2.0
+
+- Adds and enforces schema validation of ui5.yaml
+- The namespace for application and library projects is now mandatory.
```
RandomByte: This is actually Specification Version independent. Namespace is required for all projects starting with UI5 Tooling 2.0.

So this should be removed here.

docs/pages/Configuration.md
```diff
@@ -308,13 +308,14 @@ All changes are documented below.
 
 ### Compatibility Matrix
 
-Unless otherwise noted in the table below, the UI5 Tooling modules are backward compatible in the means that for example UI5 CLI v2.0 will still be able to handle a project that is using Specification Version `1.0`.
+Unless otherwise noted in the table below, the UI5 Tooling modules are backward compatible in the means that for example UI5 CLI v3.0.0 might still be able to handle a project that is using Specification Version `1.0`.
```
KlattG: I'd suggest to modify this sentence as follows:
"Unless otherwise noted in the table below, the UI5 Tooling modules are backward compatible."

RandomByte: Thanks, we'll change it to match your proposal in a separate Pull Request 👍

PR #219, opened 2020-03-26
[INTERNAL] Archive relevant v1 documentation for later reference

RandomByte:


matz3 (dismissed):

RandomByte:
I refrain from merging this since the archived documentation would appear in the normal search. Maybe we should build the legacy documentation separately in a different folder and just crosslink to it

matz3:
As it's not expected to change, we might just build it once and then copy the folder into `docs/v1` during the build?

RandomByte:
> As it's not expected to change, we might just build it once and then copy the folder into `docs/v1` during the build?

That's a good idea 👍

PR #228, opened 2020-04-16
[INTERNAL] Add "ui5 remove" cli command documentation

tobiasso85:
Depends on https://github.com/SAP/ui5-cli/pull/327

### DIFF
--- a/docs/pages/CLI.md
+++ b/docs/pages/CLI.md
@@ -27,6 +27,10 @@ Commands:
   ui5 init
       Initialize the UI5 Tooling configuration for an application or
         library project.
+
+  ui5 remove <framework-libraries..>
+      Remove SAPUI5/OpenUI5 framework libraries from the
+        project configuration.
   
   ui5 serve
       Start a web server for the current project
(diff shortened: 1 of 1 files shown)


matz3:
Looks good. To be merged once ui5-cli v2.1.0 is released.

KlattG:
How about version information?

matz3:
> How about version information?

Do you mean to add a "since" information for the commands?
I think that would be very helpful. We should do this for "use", "add" (since v2.0) and "remove" (since v2.1).

KlattG:
> > How about version information?
> 
> Do you mean to add a "since" information for the commands?
> I think that would be very helpful. We should do this for "use", "add" (since v2.0) and "remove" (since v2.1).

I mean that the CLI version is mentioned several times in the Configuration, so I'm a bit surprised that it's not mentioned here at all.

tobiasso85:
> > > How about version information?
> > 
> > 
> > Do you mean to add a "since" information for the commands?
> > I think that would be very helpful. We should do this for "use", "add" (since v2.0) and "remove" (since v2.1).
> 
> I mean that the CLI version is mentioned several times in the Configuration, so I'm a bit surprised that it's not mentioned here at all.

will be addressed in separate change

PR #230, opened 2020-04-23
[INTERNAL] Add "customConfiguration" configuration documentation

tobiasso85:
depends on https://github.com/SAP/ui5-project/pull/308

### DIFF
--- a/docs/pages/Configuration.md
+++ b/docs/pages/Configuration.md
@@ -122,6 +122,23 @@ resources:
     propertiesFileSourceEncoding: UTF-8|ISO-8859-1
 ````
 
+## Custom Properties in Configuration
+
+!!! info
+    This configuration is available since UI5 CLI [`v2.1.0`](https://github.com/SAP/ui5-cli/releases/tag/v2.1.0)
+    and applies only to projects defining [Specification Version](#specification-versions)
+    2.1 or higher.
+
+Define your project's custom configuration.
+
+In your project's custom properties you can define own attributes which you can then use in your code.
+You can define simple values or complex objects in this configuration section.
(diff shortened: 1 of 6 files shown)


docs/pages/Configuration.md
```diff
@@ -449,11 +449,20 @@ Unless otherwise noted in the table below, the UI5 Tooling modules are backward
...
 **1.0** | v1.0.0+
 **0.1** | v0.0.1+
 
+### Specification Version 2.1
+
+**Features:**
+
+- Adds support for the ["customConfiguration"](#custom-configuration) configuration to provide custom data. Must be handled by custom tools. Can be consumed when using @ui5/project API.
+
+Version 2.1 projects are supported by [UI5 CLI](https://github.com/SAP/ui5-cli) v.2.1.0 and above.
```
KlattG, line 464: What versions are these? Is Version 2.1 the UI5 Tooling version or the specVersion? And how about the UI5 CLI Release v.2.1.0 - is that yet another different version? This is not clear to me.

matz3: With `Version 2.1` the specVersion is meant.
And UI5 CLI v2.1.0 is the upcoming UI5 CLI release that supports specVersion 2.1.

We should make this a bit more clear. The text is actually taken from the previous version descriptions below, so we should update all of them.

KlattG: Would give consideration to moving the section '## Specification Versions' to the beginning, since it's such a central concept for the configuration of UI5 Tooling. It's mentioned all over the place.

matz3:
Ok from my side

PR #540, opened 2021-07-20
[INTERNAL] generateCliDoc: Fix description, add details column, fix user home path

matz3:


matz3:
WIP 🚧 
Types and default values are currently cut-off and missing

scripts/generateCliDoc.js
```diff
@@ -195,18 +190,23 @@ function generateDoc() {
...
-function splitString(command, temp, description) {
-	command = temp.split(/[^\w.,]\s(?=[A-Z])/)[0].trim();
-	description = temp.split(/[^\w.,]\s(?=[A-Z])/)[1].trim();
-	return { command, description };
+function splitString(temp) {
+	let command, description, details;
+
+	const match = temp.trim().split("  ").filter((s) => s).map((s) => s.trim());
+	if (match.length && match[match.length - 1].startsWith("[") && match[match.length - 1].endsWith("]")) {
+		details = match.splice(match.length - 1, 1)[0];
```
matz3, line 198: This just takes and removes the last element from the array, right?
'pop' does exactly that:
```suggestion
		details = match.pop();
```

scripts/generateCliDoc.js
```diff
@@ -195,18 +190,23 @@ function generateDoc() {
...
 }
 
-function splitString(command, temp, description) {
-	command = temp.split(/[^\w.,]\s(?=[A-Z])/)[0].trim();
-	description = temp.split(/[^\w.,]\s(?=[A-Z])/)[1].trim();
-	return { command, description };
+function splitString(temp) {
+	let command, description, details;
+
+	const match = temp.trim().split("  ").filter((s) => s).map((s) => s.trim());
```
matz3, line 196: Wouldn't it make more sense to first trim and then filter? Otherwise we may again end up with empty strings. Or is this expected in some cases?

scripts/generateCliDoc.js
```diff
@@ -195,18 +190,23 @@ function generateDoc() {
...
+	if (match.length && match[match.length - 1].startsWith("[") && match[match.length - 1].endsWith("]")) {
+		details = match.splice(match.length - 1, 1)[0];
+	}
+	description = match.length ? match.splice(match.length - 1, 1)[0] : "";
```
matz3, line 200:
```suggestion
	description = match.length ? match.pop() : "";
```

scripts/generateCliDoc.js
```diff
@@ -195,18 +190,23 @@ function generateDoc() {
...
+	command = match.length ? match[match.length - 1] : "";
```
matz3, line 201: Not sure if this is easier to understand in general, but IMO to me it is, and it would also align with the previous lines by always pop'ing the elements:
```suggestion
	command = match.pop() || "";
```

matz3:
@larskissel the result looks perfect to me now 👍🏻 
I've just posted some comments about the code.

And I've noticed that the path to the home folder is `$HOME//.ui5`, which I changed to `~/.ui5`, also to align with https://sap.github.io/ui5-tooling/pages/Troubleshooting/#ui5-taking-too-much-disk-space where the same path is referenced.
I didn't test that on Windows, but at least for the deployment it isn't relevant. And I'd expect that if it worked before, it will still work fine.

scripts/generateCliDoc.js
larskissel: Don't you think it is a better approach to do the filter first (many items are reduced to an array with a maximum of 3 items) and then trim the starting and ending spaces that could be left?
Otherwise we may trim a lot empty string items what would be unnecessary.

larskissel (approved):
LGTM

PR #619, opened 2022-05-13
[INTERNAL] Use JamesIves/github-pages-deploy-action for deployment

flovogt:


.github/workflows/deploy-docs.yml
```diff
@@ -25,7 +25,10 @@ jobs:
...
-      uses: maxheld83/ghpages@v0.3.0
-      env:
-        BUILD_DIR: site/
-        GH_PAT: ${{ secrets.GH_PAT }}
+      uses: JamesIves/github-pages-deploy-action@v4.3.3
+      with:
+        branch: gh-pages
+        folder: site
+        token: ${{ secrets.GH_PAT }}
+        git-config-name: OpenUI5 Bot
```
RandomByte, line 33: I wonder where this came from before? 🤔 

From the current gh-pages branch, I *think* the name and email should rather be `openui5bot` and `openui5bot@users.noreply.github.com` to match up with the actual [GitHub user](https://github.com/openui5bot): https://github.com/SAP/ui5-tooling/commit/b513dbf8ea853de702b8170301f8f67ac6325c5a.patch

.github/workflows/deploy-docs.yml
```diff
@@ -25,7 +25,10 @@ jobs:
...
+        git-config-email: openui5@sap.com
```
RandomByte, line 34:
```suggestion
        git-config-email: openui5bot@users.noreply.github.com
```

.github/workflows/deploy-docs.yml
RandomByte, line 33:
```suggestion
        git-config-name: openui5bot
```

RandomByte (changes requested):

.github/workflows/deploy-docs.yml
flovogt: I copied the config from the OpenUI5 Docs repository where we use 
```
git config user.name "OpenUI5 Bot"
git config user.email "openui5@sap.com"
```

This leads to author info `OpenUI5 Bot <openui5@sap.com>`. See here https://github.com/SAP/openui5-docs/commit/2f1b8e5db57f84f686413ff8ebb685dfd62adf80.patch

In the past this info was generated from the PAT. I guess we can just remove it and see what happens. It should automatically have the the same information as before.

RandomByte: I see! That makes sense. Actually I would be fine with that configuration then, since it seems to be working for openui5-docs. But ultimately both work fine so let's just get this merged and see whether it's working 👍

RandomByte (approved):

PR #634, opened 2022-07-11
[INTERNAL] Fix markdown tabs

matz3:
See: https://github.com/squidfunk/mkdocs-material/issues/3330


matz3:
This change needs to be cherry-picked onto `next`

PR #650, opened 2022-09-16
[INTERNAL] Transform module to ESM

matz3:
JIRA: CPOUI5FOUNDATION-525


package.json
```diff
@@ -17,36 +17,42 @@
...
 		"node": ">= 16.13.2",
 		"npm": ">= 8"
 	},
 	"scripts": {
-		"test": "npm run jsdoc-generate && npm run schema-generate",
+		"test": "npm run lint && npm run jsdoc-generate && npm run schema-generate && npm run generate-cli-doc",
+		"lint": "eslint ./",
 		"jsdoc": "npm run jsdoc-generate && open-cli site/api/index.html",
-		"jsdoc-generate": "$(node -p 'path.relative(__dirname, path.dirname(require.resolve(\"@ui5/cli\"))).startsWith(\"..\") && process.exit(1); process.exit(0)') && npm run jsdoc-generate-local || npm run jsdoc-generate-workspace",
+		"jsdoc-generate": "node --experimental-import-meta-resolve ./scripts/isWorkspace.js && npm run jsdoc-generate-workspace || npm run jsdoc-generate-local",
```
matz3, line 29: `import.meta.resolve` works with the flag, but I'm not in favour of using it here. The API might change in future node versions. Please use require.resolve instead, which should work just fine and is the same approach that we also use in other places.

flovogt: Agree. Fixed by https://github.com/SAP/ui5-tooling/pull/650/commits/519ecd25e696d3505b3a4adb4ab2c026470e3123.

scripts/isWorkspace.js
```diff
@@ -0,0 +1,22 @@
...
+	const cliPath = path.relative(__dirname, path.dirname(cPath));
+
+	// no workspace detected
+	if (!cliPath.startsWith("..")) {
+		process.exit(1);
+	}
+}
+
+main().catch((error) => {
+	console.log(error);
```
matz3, line 20: This won't be executed as "main" already calls process.exit.

AFAIK in ES modules you can also use async/await in the root context, so you don't need to define a function at all.

docs/index.md
```diff
@@ -82,19 +82,16 @@ This allows you to rely on UI5 Tooling for UI5-specific build functionality and
 All available APIs are documented in the [UI5 Tooling API Reference](https://sap.github.io/ui5-tooling/api/index.html).
 
 ```js linenums="1"
-const {normalizer} = require("@ui5/project");
-const {builder} = require("@ui5/builder");
+import {graphFromPackageDependencies} from "@ui5/project/graph";
```
matz3, line 85: Not sure whether we should also address the docs within this PR. I think this should be handled separately, as we currently don't create any docs from this branch and we still need to figure out how we want to provide docs for multiple versions. Maybe you can just move those changes into another PR.

flovogt: Thats true. I've removed the documentation part and created another branch https://github.com/SAP/ui5-tooling/tree/docs-esm which we can use in future

scripts/isWorkspace.js
flovogt: Refactored in 1c310e2.

.eslintrc.cjs
```diff
@@ -0,0 +1,50 @@
+module.exports = {
+	"parserOptions": {
+		"sourceType": "module",
+	},
+	"env": {
+		"node": true,
+		"es2022": true
```
flovogt, line 7: top level await is handled by `es2022`. If we decide to go with `es2022` we should update also the other UI5 tooling repositories. @matz3 @RandomByte whats your opinion on that one?

RandomByte: Sure, sounds good 👍

flovogt: PRs for all other repos are created

matz3:
LGTM now. @RandomByte do you want to have another look?

flovogt (approved):
LGTM. @RandomByte please have also a look on it 👀 because @matz3 and myself did the actual implementation.

RandomByte (approved):

PR #664, opened 2022-10-31
[RFC] 0012 UI5 Tooling Extension API v3

RandomByte:
Read the current draft here: **[0012-UI5-Tooling-Extension-API-3.md](https://github.com/SAP/ui5-tooling/blob/rfc-ui5-tooling-extension-api-v3/rfcs/0012-UI5-Tooling-Extension-API-3.md)**


vobu:
i like many -if not all- of the proposed enhancement! Also, introducing a breaking change w/ `v3` is fine from my POV as providing backwards compatibility at that point would mean too much technical debt to maintain.

That being said, I'm somewhat torn on the ESM-only approach of providing the ui5 tooling modules themselves as this forces migration of all tasks and middlewares requiring any `@ui5/*` modules. The CJS/ESM split has caused enough uproar in the Node-community already and we shouldn't amplify this in UI5-land. Thus, I'd suggest to provide both ESM- and CJS-modules, [similar to how fastify is doing it](https://github.com/fastify/fastify/blob/9ffb9fa38b564953208672456dd08586bedc2933/fastify.js#L715); and let the users decide on what to use.

What I'd like to also see is some improvement on bundling size - `ui5 build all` produces huge bundles that from my admittedly naïve POV could be shaken down significantly 😸

mauriciolauffer:
I agree with @vobu , especially on the build comment. Today, "tree shaking" ignores `sap.ui.core` and shakes just other modules.

mauriciolauffer:
I'd love to see the end of `ui5` in package.json.

`"ui5": {
        "dependencies": [
            "no-more!"
        ]
    }`

rfcs/0012-UI5-Tooling-Extension-API-3.md
```diff
@@ -0,0 +1,415 @@
...
+Migrating the custom task example from our current [documentation](https://sap.github.io/ui5-tooling/pages/extensibility/CustomTasks/#example-libtasksgeneratemarkdownfilesjs) to make use of the new API could look like this:
+
+```patch
+const path = require("path");
+-const {Resource} = require("@ui5/fs");
+-const log = require("@ui5/logger").getLogger("builder:customtask:markdown");
+const renderMarkdown = require("./renderMarkdown");
+
+-module.exports = async function({workspace, dependencies, taskUtil, options}) {
++module.exports = async function({workspace, dependencies, log, taskUtil, options}) {
```
flovogt, line 119: should we use here an ESM export to promote the new feature?

rfcs/0012-UI5-Tooling-Extension-API-3.md
```diff
@@ -0,0 +1,392 @@
...
+* `getVersion`: Typically returns the package.json version of the project
+* *async* `getRootReader`: Returns an `@ui5/fs/AbsatractReader` instance for the root directory of the project. Typically the same directory that contains the ui5.yaml
+* *async* `getReader`: Returns an `@ui5/fs/AbsatractReader` instance for the resources of the project
+* `getCustomConfiguration`: Returns the custom configuration of the project
+* `isFrameworkProject`: Returns true if the project is a UI5 framework (theme-)library
+
+#### Solution Example
+
+```javascript
+module.exports = async function({workspace, dependencies, taskUtil, options}) {
```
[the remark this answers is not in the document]
RandomByte: Good idea. But I think I'll rather add that to the documentation in https://github.com/SAP/ui5-tooling/pull/700

The RFC should be kept simple to illustrate the changes/described API. The documentation is probably the better place to promote the use of ESM 👍

rfcs/0012-UI5-Tooling-Extension-API-3.md
RandomByte: See https://github.com/SAP/ui5-tooling/pull/664#discussion_r1048258008

flovogt (dismissed):

flovogt (approved):

PR #697, opened 2022-11-21
[INTERNAL] Create versioning for the documentation

d3xter666:
JIRA: CPOUI5FOUNDATION-565

This change creates a new documentation release workflow that enables versioning.

It extends the MkDocs `squidfunk/mkdocs-material` image by adding [mike](https://github.com/jimporter/mike) to it.
The GitHub Actions workflow is also being extended to support the new feature.

URLs of the documentation would be changed with a version number prepending the path. In order to "keep" old links available for the end users, there's a custom redirect for GH Pages.


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,61 @@
+name: Build and Deploy Docs
+on:
+  push:
+    branches:
+      - master
```
flovogt, line 5: there is no master branch anymore. Can be removed.

scripts/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+RUN pip install mike
+
+# Overwrite ENTRYPOINT from squidfunk/mkdocs-material:8.5.9
+ENTRYPOINT [""]
+
+# Start mkdocs development server by default
+CMD [ "mkdocs", "serve", "--dev-addr=0.0.0.0:8000" ]
```
flovogt, line 9: missing empty last line

scripts/buildDocs.sh
```diff
@@ -5,8 +5,28 @@ cd "$(dirname -- "$0")/.."
...
 
-docker run --rm -it -v $(pwd):/docs squidfunk/mkdocs-material:6.0.2 build
+# Build with MkDocs/Mike
+docker run --rm -v $(pwd):/docs squidfunk/mkdocs-material:8.5.9 mike deploy $MIKE_VERSION $MIKE_ALIAS --rebase --update-aliases
 
 npm run jsdoc-generate
+
+# Set default versioning and publish
+docker run --rm -v $(pwd):/docs squidfunk/mkdocs-material:8.5.9 mike set-default $MIKE_VERSION --push
+echo "Documentation build & tagged"
```
flovogt, line 32: missing empty last line

scripts/buildDocs.sh
```diff
@@ -5,8 +5,28 @@ cd "$(dirname -- "$0")/.."
 
 echo "Changed directory to $(pwd)"
 
+# Build image if not existent
+if [[ "$(docker images -q squidfunk/mkdocs-material:8.5.9 2> /dev/null)" == "" ]]; then
+  bash ./scripts/buildImage.sh
+fi
+
+# Mike configuration
+git config --local user.email "github-actions[bot]@users.noreply.github.com"
```
flovogt, line 14: why not using our openui5 bot for this one?

scripts/buildDocs.sh
```diff
@@ -5,8 +5,28 @@ cd "$(dirname -- "$0")/.."
 
 echo "Changed directory to $(pwd)"
 
+# Build image if not existent
+if [[ "$(docker images -q squidfunk/mkdocs-material:8.5.9 2> /dev/null)" == "" ]]; then
```
flovogt, line 9: `squidfunk/mkdocs-material:8.5.9` is used several times, so might be possible to be stored in a variable so declaration only happens once

scripts/buildImage.sh
```diff
@@ -0,0 +1,11 @@
...
+set -e
+
+cd "$(dirname -- "$0")/../"
+echo "Changed directory to $(pwd)"
+
+echo "Building image..."
+docker build -t squidfunk/mkdocs-material:8.5.9 -f scripts/Dockerfile .
+echo "Done building image."
+
+exit 0
```
flovogt, line 11: missing last empty line

scripts/serveDocs.sh
```diff
@@ -7,4 +7,9 @@ echo "Changed directory to $(pwd)"
 
 npm run generate-cli-doc
 
-docker run --rm -it -p 8000:8000 -v $(pwd):/docs squidfunk/mkdocs-material:6.0.2
+# Build image if not existent
+if [[ "$(docker images -q squidfunk/mkdocs-material:8.5.9 2> /dev/null)" == "" ]]; then
```
flovogt, line 11: `squidfunk/mkdocs-material:8.5.9` is used several times, so might be possible to be stored in a variable so declaration only happens once

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+        git config --local user.email "github-actions[bot]@users.noreply.github.com"
+        git config --local user.name "github-actions[bot]"
+
+    - name: Build Mike Image
+      run: |
+        chmod +x ./scripts/buildImage.sh
+        ./scripts/buildImage.sh
+
+    - name: Build MkDocs with Mike (versioning)
+      run: docker run --rm -v $(pwd):/docs squidfunk/mkdocs-material:8.5.9 mike deploy $MIKE_VERSION $MIKE_ALIAS --rebase --update-aliases
```
flovogt, line 35: Specifying the mike version in a env variable https://docs.github.com/en/actions/learn-github-actions/environment-variables. So the version has to be maintained only once

flovogt (changes requested):

.github/workflows/deploy-docs.yml
d3xter666: I'm not quite sure I understand this one. Could you elaborate a bit more?

I think it's already been done and reused here and below. Please take a look at line #10:

>     env:
>       MIKE_VERSION: 0.9
>       MIKE_ALIAS: latest

Is this the env variable you refer to?

flovogt: Sorry I meant the version of the `squidfunk/mkdocs-material:8.5.9`. This could be stored in a nenv

d3xter666: Done

scripts/buildDocs.sh
d3xter666: Done

mkdocs.yml
```diff
@@ -42,10 +42,23 @@ theme:
...
+extra:
+  version:
+    provider: mike
+
 plugins:
   - search
   - minify:
       minify_html: true
+  - mike:
+      # These fields are all optional; the defaults are as below...
```
flovogt, line 54: comments can be removed

flovogt (changes requested):

mkdocs.yml
d3xter666: Done

.github/workflows/deploy-docs.yml
d3xter666: To be honest I'm not quite sure how to handle this. Currently, I see that the main branch does not have this workflow, although it used to have it.
My point was to leave it like this for now, so that we don't trigger build by accident as there's no master branch anymore

scripts/resources/custom404.html
```diff
@@ -0,0 +1,32 @@
...
+		url.pathname = path.join("/");
+		window.location = url.toString();
+	</script>
+</head>
+
+<body>
+	Redirecting to <a href="latest/">the latest docs release/</a>...
+</body>
+
+</html>
```
flovogt, line 32: missing empty last line

flovogt (changes requested):

scripts/resources/custom404.html
d3xter666: Done

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: GitHub Configuration
+      run: |
+        git fetch origin gh-pages --depth=1
+        git config --local user.email $GH_EMAIL
+        git config --local user.name $GH_NAME
+
+    - name: Build Mike Image
+      run: |
+        chmod +x ./scripts/buildImage.sh
```
matz3, line 35: Please do that locally, so that it can also be executed locally. A file in git also contains the execution flag.

matz3 (changes requested):

scripts/buildImage.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname -- "$0")/../"
+echo "Changed directory to $(pwd)"
+
+echo "Building image..."
+docker build -t squidfunk/mkdocs-material:8.5.9 -f scripts/Dockerfile .
```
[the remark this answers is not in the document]
d3xter666: I was thinking the same. However, it's the same image, BUT with additional plugin(s) inside.
It probably makes sense, to tag it with some modifier after all.

Thanks!

.github/workflows/deploy-docs.yml
d3xter666: Thanks a lot! Didn't know that :)

matz3: Please also remove this line then.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,61 @@
...
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    name: Build and Deploy
+    runs-on: ubuntu-20.04
+    env:
+      MIKE_VERSION: v3
+      MIKE_ALIAS: latest
+      DOCKER_IMAGE: squidfunk/mkdocs-material-modified:8.5.9
```
matz3, line 13: I still find the image name confusing. It starts with `squidfunk/` so at first look it seems to be a thirdparty image.
I'd suggest to use `ui5-tooling/mkdocs-material`.

matz3 (changes requested):

.github/workflows/deploy-docs.yml
d3xter666: What about the version? Would you prefer to keep the version aligned with the original image or to omit it and always be on the "latest"?

matz3: I think we don't need a version or tag at all, as we always re-build the image.

scripts/buildDocs.sh
```diff
@@ -27,9 +33,26 @@ fi
 
 npm run generate-cli-doc
 
+# Setup Git config for mike & stash local settings to prevent overwrites
+LOCAL_GH_NAME=$(git config --default "" --local --get user.name)
```
matz3, line 37: We only want to do this in CI, not when running the script locally. Otherwise we might create commits on behalf of OpenUI5 Bot (which happened already for this PR).

Isn't there a way to only have those settings within the GitHub workflow?

matz3: (This comment is not for this exact line, but the overall handling and (re-)setting of the git user.name/email)

d3xter666: Agree with that, but for some reason, Mike does not want to work with `git config --global` settings and for development, we need to explicitly provide `git config --local` user.name and user.email.

So, my point was, if we have set them up, then nothing happens, just the checks. But If they are not provided, to use CI defaults "for a while" and reset `git config --local` user.name and user.email. Basically, all this code is needed if we don't have `git config --local`

What do you think? Should we handle this or enforce developers to set those settings on their own?

matz3: What about using a custom docker entrypoint as bash script within our image, that ensures the git config, and then runs mike?

But still, I don't quite understand why we can't set the `git config --local` within the workflow before running the container. As long as the `.git` folder is available within the mounted volume, the config should be available in the container.

scripts/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+# ENV mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd use a specific dev version
+RUN pip install "mike>=2" || pip install -e git+https://github.com/jimporter/mike.git@2faa33ca4cf0da62c9c0ac7fbe8b60688733d7f1#egg=mike
```
matz3, line 4: This is fine for me, but I'd rather just use the commit hash version. This way we might experience issues or different behaviour in future once version 2 is released.

scripts/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+# ENV mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd use a specific dev version
```
matz3, line 3:
```suggestion
# mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd use a specific dev version
```

Please also link the GitHub issue that refers to this information.

scripts/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# Define env variables to for mike
+ENV GIT_COMMITTER_NAME="OpenUI5 Bot"
+ENV GIT_COMMITTER_EMAIL="openui5@sap.com"
+
+# Overwrite ENTRYPOINT from squidfunk/mkdocs-material:8.5.9
+ENTRYPOINT [""]
+
+# Start mkdocs development server by default
+CMD [ "mkdocs", "serve", "--dev-addr=0.0.0.0:8000" ]
```
matz3, line 14: To run mike, you could also just pass `--entrypoint mike` to override it. That way the ENTRYPOINT / CMD can stay as before (inherited from squidfunk/mkdocs-material).

scripts/buildDocs.sh
```diff
@@ -1,12 +1,37 @@
 #!/bin/bash
 set -e
 
+# Store docker image name
+DOCKER_IMAGE=ui5-tooling/mkdocs-material
+GIT_COMMITTER_NAME: "OpenUI5 Bot"
+GIT_COMMITTER_EMAIL: "openui5@sap.com"
```
matz3, line 7: this is valid bash? Never saw this syntax before

scripts/buildDocs.sh
```diff
@@ -1,12 +1,37 @@
...
+	MIKE_ALIAS=latest
+fi
+
 cd "$(dirname -- "$0")/.."
 
 echo "Changed directory to $(pwd)"
 
+# Build image if not existent
+if [[ "$(docker images -q $DOCKER_IMAGE 2> /dev/null)" == "" ]]; then
+  bash ./scripts/buildImage.sh
```
matz3, line 21: `bash` shouldn't be needed as long as the file is executable.

scripts/buildImage.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname -- "$0")/../"
```
matz3, line 4: As we don't want to add files to the image, we don't need to run within the project root.
So the docker context (i.e. cwd) can be the scripts directory.

matz3 (changes requested):

scripts/buildDocs.sh
d3xter666: No, copied it from the GH workflow and pushed a bit too early :)

scripts/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+# ENV mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd use a specific dev version
+RUN pip install "mike>=2" || pip install -e git+https://github.com/jimporter/mike.git@2faa33ca4cf0da62c9c0ac7fbe8b60688733d7f1#egg=mike
+
+# Define env variables to for mike
```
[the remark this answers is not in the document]
d3xter666: Done

scripts/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+# mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd
+# use a specific dev version: https://github.com/jimporter/mike/issues/125
+RUN cd "$(python -c 'import site; print(site.getsitepackages()[0])')" && \
```
d3xter666, line 5: We need to CD into Python's site-packages directory because for some reason pip install ignores the --target or --root flags and installs the packages into the `./src` folder.

And our current directory is the mounted one. In the case of installing into the current directory, when the container gets build, mike works fine. But when we need to mount the folder once again for another `docker run`, it fails

scripts/buildImage.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 set -e
 
-cd "$(dirname -- "$0")/../"
```
matz3: I expected to `cd` into the `scripts` directory, to be independent of the CWD location where this script is executed from.

.github/workflows/deploy-docs.yml
```diff
@@ -42,9 +42,10 @@ jobs:
       with:
         ref: gh-pages
         path: gh-pages
-    - name: Copy the API to gh-pages
+    - name: Copy the API & JSON Schem to gh-pages
```
matz3, line 45: Schema

flovogt (dismissed):

flovogt (commented):
LGTM. @matz3 and @RandomByte feel free to add some further feedback

scripts/serveDocs.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
 set -e
 
-cd "$(dirname -- "$0")/.."
```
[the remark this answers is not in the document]
d3xter666: Reordered a bit as `buildImage.sh` also CDs into some folders

matz3 (changes requested):

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,57 @@
+name: Build and Deploy Docs
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    name: Build and Deploy
+    runs-on: ubuntu-20.04
```
flovogt, line 9: why not using ubuntu lts 22.04?

matz3 (commented):
LGTM

.github/workflows/deploy-docs.yml
d3xter666: I copied it from the previous/current implementation: https://github.com/SAP/ui5-tooling/blob/v2/.github/workflows/deploy-docs.yml#L9

To be honest, didn't pay much attention to that.
Do you think we should update it? Is this 22.04 some LTS?

flovogt: keep it... will update the ubuntu version in a next step

flovogt (commented):
LGTM

RandomByte (changes requested):

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      MIKE_VERSION: v2
```
RandomByte, line 11: Shouldn't this be v3 for the main branch?

RandomByte (changes requested):

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+    - name: Install npm dependencies
+      run: npm ci
+
+    - name: GitHub Configuration
+      run: git fetch origin gh-pages --depth=1
+
+    - name: Build docs with Mike
+      run: ./scripts/buildDocs.sh
+
+    - name: Publish docs
```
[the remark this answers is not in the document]
d3xter666: It is a real `push origin gh-pages` . 
The previous step `Build docs with Mike` would build the mkdocs documentation and would commit the result to the local `gh-pages` branch. This step would update the `gh-pages/versions.json` file and enforce `$MIKE_VERSION` as the default one. The `--push` flag would push the "local" `gh-pages` branch to the `origin`

scripts/buildDocs.sh
```diff
@@ -2,11 +2,40 @@
...
 
 cd "$(dirname -- "$0")/.."
-
 echo "Changed directory to $(pwd)"
 
+# Store docker image name
+DOCKER_IMAGE=ui5-tooling/mkdocs-material
+
+# Build image if not existent
+if [[ "$(docker images -q $DOCKER_IMAGE 2> /dev/null)" == "" ]]; then
```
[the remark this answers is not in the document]
RandomByte: Ups, I moved this comment to https://github.com/SAP/ui5-tooling/pull/697#discussion_r1048249489 and forgot to delete it here!

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,57 @@
...
+      run: npm run schema-generate
+
+    - name: Checkout gh-pages
+      uses: actions/checkout@v3
+      with:
+        ref: gh-pages
+        path: gh-pages
+    - name: Copy the API & JSON Schema to gh-pages
+      run: |
+        cp -R ./site/schema ./gh-pages/ 2>/dev/null || :
```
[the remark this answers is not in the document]
d3xter666: I was wondering how to properly handle those.
It copies the results/outputs from the previous steps `Build JSDoc` & `Build Schema` to the correct locations. 
The problem is that there might be cases (at least when I run those scripts out of the context of UI5 Tooling meta repo i.e. fork of ui5-tooling poject on my own GH account) where the results of those scripts are null. Then, for example, `schema` could have changed, but the `api` might not be build. Failing any of those copies, would fail the final push to the `gh-pages` branch. And yet, there might be resources that could have been published. This is the core reasoning behind those.
Do you think thare might be a better handling for those cases? Or we could ignore entirelly the possibility not to copy everything?

scripts/Dockerfile
d3xter666: Thanks a lot!

Maybe, you're right that `python -m` is the key here. Not sure what's the difference :D

scripts/buildImage.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname -- "$0")"
+echo "Changed directory to $(pwd)"
+
+echo "Building image..."
+docker build -t ui5-tooling/mkdocs-material -f Dockerfile .
```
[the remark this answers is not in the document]
d3xter666: Regarding the caching:
- locally: yes, this would make it more convenient by tagging it with a version
- For GH Actions: It'd always be built. We do not cache it there. I'm not sure whether we should as in terms of time, it takes some negligible seconds.

Regarding the docker image existence check, I was having some thoughts in the same direction, but can't explain why didn't do it earlier 😆 

On the last point, regarding the tagging with the sample code, I'm not sure I fully grasp the idea 🤔 
`ui5-tooling/mkdocs-material` would always be `latest` and would always be (re-)build.
The `ui5-tooling/mkdocs-material:$DOCKER_TAG` would derrive from `ui5-tooling/mkdocs-material` and would be versioned with git's latest hash

I don't quite understand why we need both `ui5-tooling/mkdocs-material:$DOCKER_TAG` & `ui5-tooling/mkdocs-material`.

Which of those images/tags would we use in `serveDocs.sh`, `buildDocs.sh`, and eventually on the CI?

scripts/buildDocs.sh
```diff
@@ -2,11 +2,40 @@
...
+if [[ -z "${GIT_COMMITTER_NAME}" ]]; then
+	GIT_COMMITTER_NAME="OpenUI5 Bot"
+	GIT_COMMITTER_EMAIL="openui5@sap.com"
+fi
+
+# If not provided, set default values for building the docs
+if [[ -z "${MIKE_VERSION}" ]]; then
+	MIKE_VERSION=v3
+	MIKE_ALIAS=latest
+fi
```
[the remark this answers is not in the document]
d3xter666: Done

.github/workflows/deploy-docs.yml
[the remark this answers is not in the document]
d3xter666: Removed the error supression

scripts/buildImage.sh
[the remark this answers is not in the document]
d3xter666: We've agreed to try to find a cross-platform independent way to check whether the Docker file has been modified. That way we'd be able to determine when we need to rebuild the Docker image and when not. This scenario is quite valid for local development. 
The solution with git hash might not be that convenient as the developer must first commit before testing/executing the image building

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  push:
+    branches:
+      - main
+jobs:
+  build-and-deploy:
+    name: Build and Deploy
+    runs-on: ubuntu-20.04
+    env:
+      MIKE_VERSION: v3
```
d3xter666, line 11: For v2 branch those would need to be changed to:
`MIKE_VERSION: v2`
`MIKE_ALIAS: stable`

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,59 @@
+name: Build and Deploy Docs
```
d3xter666, line 1: We've agreed that this workflow would be copied in the v2 branch and would be slightly modified in order to fit its needs

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+      run: npm ci
+
+    - name: Fetch gh-pages branch
+      run: git fetch origin gh-pages --depth=1
+
+    - name: Build docs with Mike
+      run: ./scripts/buildDocs.sh
+
+    - name: Publish docs
+      run: docker run --rm -v $(pwd):/docs --entrypoint mike --env GIT_COMMITTER_NAME="${GIT_COMMITTER_NAME}" --env GIT_COMMITTER_EMAIL="${GIT_COMMITTER_EMAIL}" $DOCKER_IMAGE set-default stable --push
```
d3xter666, line 31: stable refers to `MIKE_ALIAS=stable` which is currently referencing the v2 branch. 
That way we'd not need to modify this line at all, but it'd always redirect to the stable/default branch

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+
+    - name: Checkout gh-pages
+      uses: actions/checkout@v3
+      with:
+        ref: gh-pages
+        path: gh-pages
+    - name: Copy the additional resources to gh-pages
+      run: |
+        rm -rf ./gh-pages/schema
+        cp -R ./site/schema ./gh-pages/
```
d3xter666, line 48: these ones need to be removed from v2

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+        rm -rf ./gh-pages/$MIKE_VERSION/api
+        cp -R ./site/api ./gh-pages/$MIKE_VERSION/
+        cp ./scripts/resources/custom404.html ./gh-pages/404.html
```
d3xter666, line 51: this one also doesn't seem to be relevant for v2

scripts/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM squidfunk/mkdocs-material:8.5.9
+
+# mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'd
```
KlattG, line 3:
```suggestion
# mike version 2 is required in order to enable ENV vars for the container. By the time mike 2 is released we'll
```

scripts/buildDocs.sh
```diff
@@ -2,11 +2,29 @@
 set -e
 
 cd "$(dirname -- "$0")/.."
-
 echo "Changed directory to $(pwd)"
 
+# Store docker image name
+DOCKER_IMAGE=ui5-tooling/mkdocs-material
+
+# Build image if not existent
```
KlattG, line 10:
```suggestion
# Build image if not existing
```

scripts/buildDocs.sh
```diff
@@ -2,11 +2,29 @@
...
+./scripts/buildImage.sh
+
 npm run generate-cli-doc
 
-docker run --rm -it -v $(pwd):/docs squidfunk/mkdocs-material:6.0.2 build
+
+if [[ $MIKE_VERSION ]]; then # Build with Mike (versioning)
+	echo "Start building & versioning Docs with Mike version: ${MIKE_VERSION}; alias: ${MIKE_ALIAS}..."
```
KlattG, line 17: Starting ?

scripts/buildDocs.sh
```diff
@@ -2,11 +2,29 @@
...
+	docker run --rm -v $(pwd):/docs --entrypoint mike \
+		--env GIT_COMMITTER_NAME="${GIT_COMMITTER_NAME}" --env GIT_COMMITTER_EMAIL="${GIT_COMMITTER_EMAIL}"  \
+		$DOCKER_IMAGE deploy $MIKE_VERSION $MIKE_ALIAS --rebase --update-aliases
+
+else # Build with MkDocs
+	echo "Start building Docs with MkDocs..."
```
KlattG, line 23: Starting ?

scripts/buildDocs.sh
```diff
@@ -2,11 +2,29 @@
...
+	docker run --rm -v $(pwd):/docs $DOCKER_IMAGE build
+
+fi
 
 npm run jsdoc-generate
+
+echo "Documentation build"
```
KlattG, line 30: Documentation has been built ?

scripts/buildImage.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname -- "$0")"
+echo "Changed directory to $(pwd)"
+
+# Store docker image name
+DOCKER_IMAGE=ui5-tooling/mkdocs-material
+# If Dockerfile has been modified, then we need to rebuild the image
```
KlattG, line 9:
```suggestion
# If Dockerfile has been modified, we need to rebuild the image
```

scripts/hash.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Utility to resolve hash of a file
```
KlattG, line 2:
```suggestion
 * Utility to resolve the hash of a file
```

scripts/resources/custom404.html
```diff
@@ -0,0 +1,32 @@
...
+	<meta charset="utf-8">
+	<title>Redirecting</title>
+	<noscript>
+		<meta http-equiv="refresh" content="1; url=latest/" />
+	</noscript>
+	<script>
+		var url = new URL(window.location);
+		var path = url.pathname.split("/");
+
+		// To prevent infinite loop redirect to the root folder if already redirected to a non-existing path.
```
KlattG, line 14:
```suggestion
		// To prevent an infinite loop, redirect to the root folder if already redirected to a non-existing path.
```

scripts/serveDocs.sh
```diff
@@ -2,9 +2,14 @@
 set -e
 
 cd "$(dirname -- "$0")/.."
-
 echo "Changed directory to $(pwd)"
 
+# Store docker image name
+DOCKER_IMAGE=ui5-tooling/mkdocs-material
+
+# Build image if not existent
```
KlattG, line 10:
```suggestion
# Build image if not existing
```

KlattG (changes requested):

scripts/buildDocs.sh
d3xter666: Thanks

scripts/buildDocs.sh
d3xter666: Thanks

KlattG (approved):
LGTM

.github/workflows/deploy-docs.yml
RandomByte: To clarify: We never want to deploy the v2 schema (does not include v3) but always the latest (includes all versions)

scripts/hash.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Utility to resolve the hash of a file
```
RandomByte, line 2: Awesome 👌

RandomByte (approved):
Looks great. But your call whether you want to try and get this working before the weekend 😜

PR #703, opened 2022-12-02
[INTERNAL] Security audit for non-default branches

d3xter666:
JIRA: CPOUI5FOUNDATION-572

Schedule a security audit on non-default branches.

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -1,0 +1,28 @@
+name: Security Audit
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: "1 1 */1 * *" # Run once a day
+
+jobs:
+  security-scan:
+    name: Security Audit
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        branch: ["v2"] # List of branches to run the security audit uppon
+
+    steps:
+      - name: Checkout '${{ matrix.branch }}' branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ matrix.branch }}
+
+      - name: install dependencies
+        run: npm ci
+
+      - name: Use audit-ci
+        run: npx audit-ci@^6 --config ./audit-ci.jsonc
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
   schedule:
-    - cron: "1 1 */1 * *" # Run once a day
+    - cron: "36 4 * * *" # Run once a day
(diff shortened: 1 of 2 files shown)


.github/workflows/security-audit.yml
```diff
@@ -0,0 +1,28 @@
+name: Security Audit
+
+on:
+  push:
+    branches:
+      - main
```
matz3, line 6: Do we really want to also trigger this on every push to main?
I think the daily schedule should be sufficient.

.github/workflows/security-audit.yml
```diff
@@ -0,0 +1,28 @@
...
+  schedule:
+    - cron: "1 1 */1 * *" # Run once a day
```
matz3, line 8: What's the advantage of `*/1` vs `*` for the day?

GitHub suggests to not start jobs at the hour. In this case it's 1 minute after, where the load is probably still high.

matz3 (changes requested):

.github/workflows/security-audit.yml
flovogt: i would suggest to run these checks early in the morning european time 04:00, or so.

.github/workflows/security-audit.yml
flovogt: we plan to have no pushes to master so often... Therefore I guess it is not such painful to have it like that

d3xter666: It was needed in order to register the workflow. But I guess as we're on the default branch, it should be redundant. I'll test to remove it

flovogt (commented):
LGTM. Waiting for @RandomByte and @matz3

audit-ci.jsonc
```diff
@@ -0,0 +1,7 @@
+{
+	// $schema provides code completion hints to IDEs.
+	"$schema": "https://github.com/IBM/audit-ci/raw/main/docs/schema.json",
+	"moderate": true,
```
matz3, line 4: I'd prefer to use `low: true` to also get informed about all vulnerabilities, also with prio low.

d3xter666: That would mean that the job would fail on any found vulnerability. This is the threshold of failing the job. 
Still, regardless of the threshold, a summary with all the findings could be found within the job.

So, the question is more like "**_Do we want to fail the job if any kind of a security vulnerability is found?_**"
If we set the threshold to `low` it would mean that we'd need to maintain (eventually) a bigger list of exceptions as you're probably aware that many of the low severity issues are most likely not to be resolved sooner or later

matz3 (approved):

PR #706, opened 2022-12-07
[BUGFIX] Security audit for non-default branches

d3xter666:
The configuration of the security audit tool is contained within the main (default) branch, but the security scan should be made on some other branch.

We need to execute the scan on a "random" branch, but take the configuration from the default one. 
So, the workflow is:
- Checkout the default branch
- Checkout the branch (in a separate folder) to which the security scan would be run & install its npm dependencies
- Run the security audit with the configuration from the main branch.

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Checkout the default branch (security configuration)
         uses: actions/checkout@v3
+        with:
+          path: default
 
       - name: Checkout '${{ matrix.branch }}' branch
         uses: actions/checkout@v3
@@ -30,4 +32,4 @@ jobs:
       - name: Security check for '${{ matrix.branch }}' branch
         run: |
           cd ${{ matrix.branch }}
-          npx audit-ci@^6 --config ../audit-ci.jsonc
+          npx audit-ci@^6 --config ../default/audit-ci.jsonc
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -13,13 +13,21 @@ jobs:
         branch: ["v2"] # List of branches to run the security audit uppon
 
     steps:
+      - name: Checkout the default branch (security configuration)
+        uses: actions/checkout@v3
+
       - name: Checkout '${{ matrix.branch }}' branch
         uses: actions/checkout@v3
         with:
           ref: ${{ matrix.branch }}
+          path: ${{ matrix.branch }}
 
-      - name: install dependencies
-        run: npm ci
+      - name: Install '${{ matrix.branch }}' dependencies
+        run: |
+          cd ${{ matrix.branch }}
+          npm ci
 
-      - name: Use audit-ci
-        run: npx audit-ci@^6 --config ./audit-ci.jsonc
+      - name: Security check for '${{ matrix.branch }}' branch
+        run: |
+          cd ${{ matrix.branch }}
+          npx audit-ci@^6 --config ../audit-ci.jsonc


.github/workflows/security-audit.yml
```diff
@@ -13,13 +13,21 @@ jobs:
...
 
     steps:
+      - name: Checkout the default branch (security configuration)
+        uses: actions/checkout@v3
+
       - name: Checkout '${{ matrix.branch }}' branch
         uses: actions/checkout@v3
         with:
           ref: ${{ matrix.branch }}
+          path: ${{ matrix.branch }}
```
RandomByte, line 23: Will this path be located inside the checked-out default branch? In that case, I think this could lead to issues when the branch name conflicts with an existing directory in the repository.

Maybe we could configure a sub-directory for the initial checkout too? Something like "default"?

RandomByte:
Regarding the commit message, please use `[FIX]` instead of `[BUGFIX]`. Also see https://github.com/SAP/ui5-tooling/blob/main/docs/Guidelines.md#commit-message-style

RandomByte:
After thinking about it for a little longer, this would mean that if we want to ignore a dependency on the v2 branch, we need to maintain this in the configuration on main? Maybe we should rather maintain a dedicated configuration file for v2. The content is rather straight forward I think.

.github/workflows/security-audit.yml
d3xter666: Thanks!

d3xter666:
@RandomByte that was my initial point, but in that case, we might end up in a situation where we need to maintain a bigger set of configurations that are different per each branch.

With the current approach, this is maintained centrally and IMO it could be better handled.

RandomByte:
> the obvious downside is that the exceptions list in the configuration could be not fully relevant for all of the branches.

Even worse, if we ignore a dependency centrally because we can't upgrade it in one branch (for example because Node.js version support limitations), we would not receive warnings for other issues with that dependencies on branches where an upgrade would be feasible.

> we could eventually check for a local config and if not, use the one in the default branch.

That also came to my find. But realistically speaking, we only need this job for the `v2` branch, and it seems natural to maintain the configuration there. I think I would expect *no* configuration file on the `main` branch (since it is currently not relevant there) and to fail the job if no configuration can be found for a given branch. Just to keep things simple until we see a need for scenarios where configuration from one branch should be used in another.

RandomByte:
I'm also interested in @matz3's view on this. Maybe we can put this PR on hold until we come to a good solution and also Matthias was able to add his thoughts?

d3xter666:
> > the obvious downside is that the exceptions list in the configuration could be not fully relevant for all of the branches.
> 
> Even worse, if we ignore a dependency centrally because we can't upgrade it in one branch (for example because Node.js version support limitations), we would not receive warnings for other issues with that dependencies on branches where an upgrade would be feasible.
> 
> > we could eventually check for a local config and if not, use the one in the default branch.
> 
> That also came to my find. But realistically speaking, we only need this job for the `v2` branch, and it seems natural to maintain the configuration there. I think I would expect _no_ configuration file on the `main` branch (since it is currently not relevant there) and to fail the job if no configuration can be found for a given branch. Just to keep things simple until we see a need for scenarios where configuration from one branch should be used in another.

I'll add just some notes here. Probably it's a bit more complicated scenario, but some CIs use special branches for that purpose. In those branches is the whole configuration for the CI. I doubt that we have such a complex scenario as you mention that we need this only for the `v2` branch and it's not worth the efforts & the complexity

.github/workflows/security-audit.yml
```diff
@@ -13,13 +13,23 @@ jobs:
         branch: ["v2"] # List of branches to run the security audit uppon
 
     steps:
+      - name: Checkout the default branch (security configuration)
+        uses: actions/checkout@v3
+        with:
+          path: default
```
matz3, line 19: What about using the actual branch name `main`?

.github/workflows/security-audit.yml
```diff
@@ -13,13 +13,23 @@ jobs:
...
+        run: |
+          cd ${{ matrix.branch }}
+          npm ci
 
-      - name: Use audit-ci
-        run: npx audit-ci@^6 --config ./audit-ci.jsonc
+      - name: Security check for '${{ matrix.branch }}' branch
+        run: |
+          cd ${{ matrix.branch }}
+          npx audit-ci@^6 --config ../default/audit-ci.jsonc
```
matz3, line 35: I would expect that the audit-ci.jsonc is stored in the branch that is tested. Otherwise we can't scale this solution to other branches. For example, I don't see a reason why we shouldn't also use this approach for the `main` branch.

d3xter666:
Abandoning this pull request as we've agreed to keep the security audit config within the scanned branch.
From that perspective this pull request becomes redundant.

PR #717, opened 2022-12-19
[INTERNAL] Docs versioning alignments

d3xter666:
This change
- Aligns mkdocs configuration between the main and v2 branches- they need to be the same.
- Cleans up whitespace as the original mkdocs.yaml
- Returns back the custom footer of the documentation (copied from v2 branch).
- bugfixes `isWorkspace.js` script to detect properly workspaces

JIRA: CPOUI5FOUNDATION-565


scripts/isWorkspace.js
```diff
@@ -10,6 +10,6 @@ const cPath = require.resolve("@ui5/cli/package.json");
 const cliPath = path.relative(__dirname, path.dirname(cPath));
 
 // no workspace detected
-if (!cliPath.startsWith("..")) {
+if (!cliPath.startsWith("../../")) {
```
RandomByte, line 13: Maybe better to use path.join to make sure this works on Windows too:

```suggestion
if (!cliPath.startsWith(path.join("..", ".."))) {
```

RandomByte (changes requested):

RandomByte (dismissed):

RandomByte (approved):

PR #720, opened 2022-12-19
[INTERNAL] Bugfix: Keep gh-pages history for mike

d3xter666:
Adds additional, not `mike`/`mkdocs` related documents with a separate commit message instead of squashing the previous one. That way the git history is kept.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -56,5 +56,5 @@ jobs:
         git config --local user.email $GIT_COMMITTER_EMAIL
         git config --local user.name $GIT_COMMITTER_NAME
         git add .
-        git commit --amend --no-edit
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -56,5 +56,5 @@ jobs:
         git config --local user.email $GIT_COMMITTER_EMAIL
         git config --local user.name $GIT_COMMITTER_NAME
         git add .
-        git commit --amend --no-edit
+        git commit -m "Updating supplemental resources for ${MIKE_VERSION} documentation deployment"
         git push --force
```
RandomByte, line 60:
```suggestion
        git push
```

Hopefully no `--force` necessary

RandomByte (changes requested):

RandomByte (approved):
Please use Squash and merge and prefix the commit with [INTERNAL]

PR #723, opened 2022-12-21
[INTERNAL] Update Documentation to SAP Web Presence Policy

flovogt:
This re-enables https://github.com/SAP/ui5-tooling/pull/618 for V3 documentation.


docs/pages/Builder.md
```diff
@@ -48,38 +48,32 @@ A project can add custom tasks to the build by using the [Custom Tasks Extensibi
...
-| generateCachebusterInfo | {: .sap-icon-circle-task } |  |  |
-| generateApiIndex | {: .sap-icon-circle-task }^1^ |  |  |
-| generateResourcesJson | {: .sap-icon-circle-task } | {: .sap-icon-circle-task } | {: .sap-icon-circle-task } |
-
-*Enabled by default*
-{: .sap-icon-circle-task-2-before }
-
-*Disabled by default. Can be activated by certain build modes, project configuration or by using the `--include-task` [CLI parameter](./CLI.md#ui5-build). See footnotes where given* 
-{: .sap-icon-circle-task-before }
+All available standard tasks are documented [in the API reference](../../api/index.html). Search for `@ui5/builder/tasks/` to filter the API reference for all available tasks. The list below offers the actual order of their execution:
```
RandomByte, line 51: The problem with the API reference is that it's not available when *serving* the documentation locally (only when building). AFAIK that's the reason we used absolute links to GitHub pages in the past. So not sure whether we want to change this.

Maybe we could come up with a script as part of the documentation build that replaces `https://sap.github.io/ui5-tooling/` with a relative URL 🤔 Might be tricky since the amount of "../" depends on the page configured in the mkdocs.yml (not the physical location of the file)

RandomByte (approved):

PR #736, opened 2022-12-27
[INTERNAL] Update usages of CJS require/exports to ESM import/export

flovogt:
Update the code samples to use additionally ESM import and export statements. The respective code blocks are displayed in tabs "CommonJS" and "ESM".

Screenshot 2023-01-02 at 12 34 28
Screenshot 2023-01-02 at 12 34 35

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -86,12 +86,12 @@ This allows you to rely on UI5 Tooling for UI5-specific build functionality and 
 
 All available APIs are documented in the [UI5 Tooling API Reference](https://sap.github.io/ui5-tooling/v3/api/index.html).
 
-=== "CommonJS"
+=== "ESM"
 
     ```js linenums="1"
+    import {graphFromPackageDependencies} from "@ui5/project/graph";
+
     async function buildApp(projectPath, destinationPath) {
-        const {graphFromPackageDependencies} = 
-            await import("@ui5/project/graph");
         const graph = await graphFromPackageDependencies({
             cwd: projectPath
         });
@@ -104,12 +104,12 @@ All available APIs are documented in the [UI5 Tooling API Reference](https://sap
     }
     ```
 
-=== "ESM"
+=== "CommonJS"
 
     ```js linenums="1"
-    import {graphFromPackageDependencies} from "@ui5/project/graph";
-
     async function buildApp(projectPath, destinationPath) {
+        const {graphFromPackageDependencies} = 
+            await import("@ui5/project/graph");
         const graph = await graphFromPackageDependencies({
             cwd: projectPath
         });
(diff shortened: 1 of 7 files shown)


docs/index.md
```diff
@@ -82,19 +82,17 @@ This allows you to rely on UI5 Tooling for UI5-specific build functionality and
 All available APIs are documented in the [UI5 Tooling API Reference](https://sap.github.io/ui5-tooling/v3/api/index.html).
 
 ```js linenums="1"
-const {normalizer} = require("@ui5/project");
-const {builder} = require("@ui5/builder");
+import {graphFromPackageDependencies} from "@ui5/project/graph";
```
RandomByte, line 85: I think it's fair to assume that many developers still work with CommonJS (just like we do with many of our internal scripts). Maybe we can put both flavors into respective [content tabs](https://squidfunk.github.io/mkdocs-material/reference/annotations/#in-content-tabs) (e.g. `ESM` and `CommonJS`)?

Same for the custom task/middleware examples.

Code looks fine 👍

RandomByte (changes requested):

docs/index.md
flovogt: Good idea! Will do that! And also provide a sample code in https://github.com/SAP/openui5-sample-app/tree/demo-server-middleware-extensibility-v3

flovogt: Done

matz3: But then we should still put **ESM** as the first tab, right?

flovogt: Good point. I'll rearrange the samples

flovogt: Done

docs/pages/TasksAndProcessors.md
```diff
@@ -1,57 +1,213 @@
 # Tasks and Processors
```
matz3, line 1: I feel like I have asked this already in the past: Is there a reason why this page is not linked anywhere?

matz3: I also think that the examples we have in the existing documentation pages are sufficient and also better, so I think we can remove this page.

matz3 (approved):

docs/pages/TasksAndProcessors.md
RandomByte: Yes, AFAIK we wanted to remove this page already in the past but linked to it from somewhere.

flovogt: Fine. I'll create a follow-up PR removing this page

flovogt: PR https://github.com/SAP/ui5-tooling/pull/742 covers this

PR #745, opened 2023-01-17
[BREAKING] Drop support for node 16

flovogt:
This PR drops support for Node 16. Only node versions higher than `18.0.0` are supported.


d3xter666:
LGTM

flovogt:
Agreed on dropping support of node version 16 in `ui5-tooling` only because devDependency has dropped node 16 support. https://github.com/SAP/ui5-tooling/pull/748

RandomByte:
I think for the `@apidevtools/json-schema-ref-parser` dependency, we also need to update Node.js in the [deploy-docs.yml](https://github.com/SAP/ui5-tooling/blob/main/.github/workflows/deploy-docs.yml). Maybe also update the ubuntu version there to 22.04

RandomByte (approved):

PR #747, opened 2023-01-18
[INTERNAL] Fix infinite looping for unknown sources

d3xter666:
This is an issue with the Docs. With the implementation of versioned documentation, we wanted to support also the legacy links. That's why, when a resource is not found we check whether this resource is in the versioned documentation and if not, we redirect to the same path of the stable release. If the resource is already versioned, but is not found, we just redirect to the stable release's root path.

### DIFF
--- a/scripts/resources/custom404.html
+++ b/scripts/resources/custom404.html
@@ -13,7 +13,7 @@
 
 			var url = new URL(window.location);
 			var path = url.pathname.split("/");
-			var pathRegex = /\/ui5-tooling\/(v[1-9]+|stable|next)/gi
+			var pathRegex = /\/ui5-tooling\/(v[0-9]+|stable|next)/gi
 
 			// For "old" links- keep the path but try to redirect to the stable release
 			if (!url.pathname.match(pathRegex)) {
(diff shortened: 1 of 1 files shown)


d3xter666:
> > If the resource is already versioned, but is not found, we just redirect to the stable release's root path.
> 
> Does this mean we would never show a 404 page?

Exactly, the scenario you're referring to is when we try to first resolve the old path by pretending the `stable` alias, but no resource has been found. 
Otherwise, we'd need to redirect not to the root path, but to version specific's 404 page. The reasoning is that GitHub seems not to handle folder-specific 404s, but just the one that's at the root of the branch.

What do you think? Would it be better then to redirect to /stable/404.html?

d3xter666:
As GitHub understands only the root folder's 404 pages and not the versioned ones i.e. `/v2/404.html`, we've agreed that if a resource is not found, then we'd redirect to the versioned 404 through the root's 404 logic.
As we cannot be 100% sure which versions/folders are physically there (i.e. `/ui5-tooling/v42389723` would match the pattern for versioning), there's an edge case when a versioned 404 is not found, to redirect to `/stable/404`

RandomByte (approved):

PR #750, opened 2023-01-24
[INTERNAL] Use specVersion 3.0 in all examples

RandomByte:


docs/index.md
```diff
@@ -29,7 +29,7 @@ Configure your project for use with UI5 Tooling.
 ❯ ui5 init
 Wrote ui5.yaml:
 
-specVersion: '2.6'
+specVersion: '3.0'
```
flovogt, line 32: should we use here also double quotes?

RandomByte: IIRC that's how `ui5 init` does it because of the library we use for serializing to YAML 🤷‍♂️

flovogt: Ack

flovogt (approved):

PR #762, opened 2023-02-07
[INTERNAL] audit-ci(v2): Add security alert for taffdb 'GHSA-mxhp-79qh-mcx6' to allow list

flovogt:
Taffdb is only used during jsdoc build and contains only data for jsdoc generation. After jsdoc build db is destroyed.


d3xter666 (approved):

ecker:
@flovogt, there's a typo "taffdb", so better fix that for "taffydb" (in issue title and code comment), so that it's easily found later.

PR #783, opened 2023-02-15
[INTERNAL] CustomTask: Fix order in JSDoc for 'determineRequiredDependencies' callback

RandomByte:
Order of parameters should be identical in ESM and CommonJS examples


docs/pages/extensibility/CustomTasks.md
```diff
@@ -254,15 +254,15 @@ If this callback is not provided, UI5 Tooling will make an assumption as to whet
      * @param {Set} parameters.availableDependencies
      *      Set containing the names of all direct dependencies of
      *      the project currently being built.
+     * @param {function} parameters.getProject
```
flovogt, line 257: why not the other way round? `P` is after `D` in the alphabet

PR #790, opened 2023-02-24
[INTERNAL] Builder.md: Mark task generateVersionInfo as enabled for jsdoc build

flovogt:
See: https://github.com/SAP/ui5-project/pull/583

### DIFF
--- a/docs/pages/Builder.md
+++ b/docs/pages/Builder.md
@@ -68,7 +68,7 @@ All available standard tasks are documented [in the API reference](https://sap.g
 | generateBundle            | *disabled* ^4^     | *disabled* ^4^ |                      |
 | buildThemes               |                    | *enabled*      | *enabled*            |
 | generateThemeDesignerResources |               | *disabled* ^5^ | *disabled* ^5^       |
-| generateVersionInfo       | *disabled* ^1^     | *disabled* ^1^ | *disabled* ^1^       |
+| generateVersionInfo       | *disabled* ^1^     |                |                      |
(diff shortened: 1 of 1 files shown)


docs/pages/Builder.md
```diff
@@ -68,7 +68,7 @@ All available standard tasks are documented [in the API reference](https://sap.g
 | generateBundle            | *disabled* ^4^     | *disabled* ^4^ |                      |
 | buildThemes               |                    | *enabled*      | *enabled*            |
 | generateThemeDesignerResources |               | *disabled* ^5^ | *disabled* ^5^       |
-| generateVersionInfo       | *disabled*         |                |                      |
+| generateVersionInfo       |                    | *disabled* ^1^ |                      |
```
matz3, line 71: I'd expect this to be in the "application" column, not in the library column, right?

matz3 (changes requested):

docs/pages/Builder.md
flovogt: I have added the term "disabled" to all columns because its relevant for any jsdoc build triggered independent of the kind of project, isn't?

matz3: Not exactly. It can't be activated in other project types, so it shouldn't be listed there. Just for application with the ^1^ remark.

flovogt: Oh I was not aware of that. Sure. will change that

matz3 (approved):

PR #792, opened 2023-02-24
[INTERNAL] csp: Update documentation for sap-csp-policies target level increase

flovogt:
See: https://github.com/SAP/ui5-server/pull/580

### DIFF
--- a/docs/pages/Server.md
+++ b/docs/pages/Server.md
@@ -33,7 +33,7 @@ E.g. `/index.html?sap-ui-xx-csp-policy=sap-target-level-1:report-only`
 
 #### The SAPtargetCSP parameter
 The default CSP policies can be modified using parameter `sendSAPTargetCSP` (`--sap-csp-policies` when using the CLI).
-With `sendSAPTargetCSP` set to `true` the policies `sap-target-level-1` and `sap-target-level-2` policies are activated and send as report-only.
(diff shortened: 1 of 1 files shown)


docs/pages/Server.md
```diff
@@ -33,7 +33,7 @@ E.g. `/index.html?sap-ui-xx-csp-policy=sap-target-level-1:report-only`
 
 #### The SAPtargetCSP parameter
 The default CSP policies can be modified using parameter `sendSAPTargetCSP` (`--sap-csp-policies` when using the CLI).
-With `sendSAPTargetCSP` set to `true` the policies `sap-target-level-1` and `sap-target-level-2` policies are activated and send as report-only.
+With `sendSAPTargetCSP` set to `true` the policies `sap-target-level-1` and `sap-target-level-3` policies are activated and send as report-only.
```
codeworrior, line 36: two comments from my side regarding the text outside your change:

- the title says `SAPtargetCSP` parameter, but there's no such parameter, it's `sendSAPtargetCSP` (from a CLI user's perspective)
- the next paragraph below says
     !!! info This option is available since UI5 CLI [v2.3.0](https://github.com/SAP/ui5-cli/releases/tag/v2.3.0)

Can't we remove that hint in version 3? I think it was necessary in v2, but v3 it doesn't make much sense to me.

matz3: No, I think we currently don't have explanation for the policies. They need to be checked in the code.
I addressed the other two points.

PR #795, opened 2023-02-28
[INTERNAL] docs: Fix link to 'Local Dependency Resolution' RFC

matz3:
Fixes: https://github.com/SAP/ui5-tooling/issues/794


ecker (approved):
Good catch, can confirm this from a pending change of mine, so let's get in your fix.

PR #813, opened 2023-03-30
[INTERNAL] Adjust occurences of outdated build information

flovogt:
Adjusts sections talking about the ui5 build.

Fixes https://github.com/SAP/ui5-tooling/issues/808.


dfenerski:
Sorry for bothering still, but I believe the `yaml` schema has to be updated too [here](https://sap.github.io/ui5-tooling/schema/ui5.yaml.json). Otherwise VSCode complains 
image
and the ui5 cli rejects it 
image

Otherwise good to know that task `settings` do exist - didn't knew that when creating one of the issues. It would be nice to be able to see which task accepts what `settings`.

Sorry again for bothering.

PR #840, opened 2023-06-27
[INTERNAL] ESSupport.md: Add a section about reserved variable names

flovogt:
List the reserved variable names inside Template Literal expressions.

### DIFF
--- a/docs/pages/ESSupport.md
+++ b/docs/pages/ESSupport.md
@@ -169,7 +169,7 @@ A library is typically initialized via an accompanying `library.js`. Within that
 
 #### Reserved Variable Names in a Template Literal
 
-Template Literals with an expression requires to relinquish following variable names:
+While UI5 Tooling performs a build placeholders are replaced with a values offered by the build. For example `${version}` is replaced with the actual version defined in the package.json of the project. Thereore it is required to not use any **Template Literal** where any expression contains variable with folloing names:
 
 - `version`
 - `project.version`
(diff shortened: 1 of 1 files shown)


docs/pages/ESSupport.md
```diff
@@ -167,6 +167,30 @@ A library is typically initialized via an accompanying `library.js`. Within that
     });
     ```
 
+#### Reserved Variable Names in a Template Literal
+
+Template Literals with an expression requires to relinquish following variable names:
```
RandomByte, line 172: From the top of my head, I did not know the meaning of `relinquish` 😄 

How about:

```suggestion
Template Literals with an embedded expression must not use the following variable names:
```

RandomByte (changes requested):

docs/pages/ESSupport.md
matz3: The reason for this limitation should be explained. We could even suggest to use https://eslint.org/docs/latest/rules/template-curly-spacing with option "always" to mitigate this. Our placeholder replacement does not allow spaces.

flovogt: Tackled both comments.

docs/pages/ESSupport.md
```diff
@@ -167,6 +167,31 @@ A library is typically initialized via an accompanying `library.js`. Within that
     });
     ```
 
+#### Reserved Variable Names in a Template Literal
+
+While UI5 Tooling performs a build placeholders are replaced with a values offered by the build. For example `${version}` is replaced with the actual version defined in the package.json of the project. Thereore it is required to not use any **Template Literal** where any expression contains variable with folloing names:
```
matz3, line 172:
```suggestion
While UI5 Tooling performs a build placeholders are replaced with a values offered by the build. For example `${version}` is replaced with the actual version defined in the package.json of the project. Therefore it is required to not use any **Template Literal** where any expression contains variable with following names:
```

matz3 (approved):

PR #844, opened 2023-07-03
[INTERNAL] Disable security audit warning for legacy semver

d3xter666:


audit-ci.jsonc
```diff
@@ -18,6 +18,12 @@
...
 		// "cacheable-request" has a dependency to "http-cache-semantics" (GHSA-rc47-6667-2j5j) which is 
 		// why it is considered as high severity. Not applicable as described above for GHSA-rc47-6667-2j5j.
-		"GHSA-8x6c-cv3v-vp6g"
+		"GHSA-8x6c-cv3v-vp6g",
+		
+		// "semver" vulnerable to Regular Expression Denial of Service.
+		// "semver" is a dependency of "make-dir" that's only used in v2 branch. As we have decided to
+		// deprecate the v2 branch and encourage people to migrate their projects to v3, we are not
+		// considering fix for this.
+		"GHSA-c2qf-rxjj-qqgw"
```
flovogt, line 27: could we limit the restriction to "make-dir" only?

d3xter666: I might try to use wildcards. Let's see

flovogt: according to https://github.com/IBM/audit-ci you can specifiy the path. 
`GHSA-c2qf-rxjj-qqgw|make-dir>semver`, I am not sure if make-dir and semver has to be reordered

d3xter666: I guess only that might not work, because this pattern appears also in other paths. According to the documentation, the same vector with different path would also fail. So, they suggest wildcards.

I have modified the code to address those. Let's see whether it would work 😄

flovogt: Ack

flovogt (approved):

PR #847, opened 2023-07-11
[INTERNAL] Disable security audit warning for legacy semver

d3xter666:


audit-ci.jsonc
```diff
@@ -28,6 +28,9 @@
 		"GHSA-c2qf-rxjj-qqgw|*make-dir>semver*",
 		"GHSA-c2qf-rxjj-qqgw|*>normalize-package-data>semver*",
 		"GHSA-c2qf-rxjj-qqgw|*npm-package-arg>semver*",
+
+		// All versions of the package word-wrap are vulnerable to Regular Expression Denial of Service (ReDoS)
```
matz3, line 32: I don't really see a benefit in repeating the advisory text here. Should we rather describe whether and how the vulnerability affects UI5 Tooling?

matz3: AFAICS https://github.com/SAP/ui5-tooling/actions/runs/5515868637 only complained about semver on top-level (`GHSA-c2qf-rxjj-qqgw|semver`).

d3xter666: Indeed, I got it from some mess on my local dev setup. Now should be fine

matz3 (approved):

PR #897, opened 2023-11-10
[RFC] 0014 Task Workers

RandomByte:
Concept for a new API provided to UI5 Tooling build tasks, enabling easy use of Node.js [Worker Threads](https://nodejs.org/api/worker_threads.html) to execute CPU intensive operations outside of the main thread.

**[View rendered version](https://github.com/SAP/ui5-tooling/blob/rfc-task-workers/rfcs/0014-task-workers.md)**

**PoC**: https://github.com/SAP/ui5-builder/pull/955
JIRA: CPOUI5FOUNDATION-751


rfcs/0014-task-workers.md
```diff
@@ -161,10 +162,24 @@ metadata:
...
+    	  path: lib/tasks/piProcessor.js
+
+
+ 	# Option 3
+    processors:
+    	computePi:
+    	  path: lib/tasks/piProcessor.js
 ```
 
+**TODO:** Decide on configuration style. Option 1 does not allow for introducing additional per-processor configuration in the future (e.g. max CPU threads, priority, etc.)
```
matz3, line 182: I find option 3 the most consistent one, as it has a very similar structure as the task itself, and order does not matter (right?), so there's no reason for option 2.

rfcs/0014-task-workers.md
```diff
@@ -177,20 +192,23 @@ The `processors.execute` function shall accept the following parameters:
...
+ * @param {@ui5/project/build/helpers/TaskUtil|object} [parameters.taskUtil] TaskUtil
+ * @param {object} processors
  * @param {Object} parameters.options Options
  * @param {string} parameters.options.projectName Project name
  * @param {string} [parameters.options.configuration] Task configuration if given in ui5.yaml
- * @param {object} processors
  * @returns {Promise<undefined>} Promise resolving with undefined once data has been written
  */
-module.exports = function({workspace, options, processors}) {
+module.exports = function({workspace, taskUtil, processors, options}) {
```
matz3, line 212: This file can also be an ES module, right?

RandomByte: Sure! Updated 👍

rfcs/0014-task-workers.md
RandomByte: I agree and updated that part

rfcs/0014-task-workers.md
```diff
@@ -0,0 +1,274 @@
...
+    - The work dispatcher shall dispatch the workload in a first in, first out order
+    - Task processors can finish in any order, and the result is supplied to the task immediately. Therefore a task processor might finish before or after another one that has been requested at a later time.
+* A single Worker shall never execute more than one task processor at a time
+* A task processors must be stateless
+
+### Assumptions
+
+* A task processor is assumed to utilize a single CPU thread by 90-100%
+* A task processor is assumed to execute little to no I/O operations
+* A task processor is assumed to make little use of UI5 Tooling modules other than those provided directly to the task processor
```
d3xter666, line 67: I would go even stricter on this one as the processor (eventually) runs in a thread and it must be totally decoupled from anything else that is not provided as an argument. It's just for the simple reason that even if it tries, it cannot simply access the resources from the main thread like this

rfcs/0014-task-workers.md
```diff
@@ -0,0 +1,274 @@
...
+### Task Processor
+
+[Processors](https://sap.github.io/ui5-tooling/stable/pages/Builder/#processors) are an established concept in UI5 Tooling but not yet exposed to custom tasks. The basic idea is that tasks act as the glue code that connects a more generic processor to UI5 Tooling. For example, UI5 Tooling processors make use of very little UI5 Tooling API, making them easily re-usable in different environments like plain Node.js scripts.
+
+With this RFC, we extend this concept to custom tasks. A task can define one or more processors and execute them with a defined API. Their execution is managed by UI5 Tooling, which might execute them on the main thread or in a worker.
+
+#### Input Parameters
+
+* **`resources`**: An array of `@ui5/fs/Resource` provided by the task
+* **`options`**: An object provided by the task
```
d3xter666, line 78: It must be a serializeable object. Otherwise, when we pass this object to the thread, some information from it could be lost. Also, it would be very painful for debugging

rfcs/0014-task-workers.md
```diff
@@ -0,0 +1,274 @@
...
+```js
+// Option 4
+export const processors = {
+    computePi: {
+        path: "../piProcessor.js"
+    }
+};
+```
+
+Option 1 does not allow for introducing additional per-processor configuration in the future (e.g. max CPU threads, priority, etc.). Option 4 can't make use of the existing schema validation of ui5.yaml-based configuration.
```
d3xter666, line 198: Maybe, I'm missing something, but if we take the PoC with `workerpool`, then workers configuration must be in a single, central place, but not within the definition of the processor. We create a singleton instance of the `workerpool` that is then reused accross the modules / tasks and the configuration of that `workerpool` happens at the first invocation of the instance.

From that point, maybe it's not feasible to provide such flexibility per processor

PR #898, opened 2023-11-24
[RFC] 0015 Build Phases

RandomByte:


rfcs/0015-build-phases.md
```diff
@@ -0,0 +1,130 @@
...
+      # No scheduling configuration necessary (although possible, overwriting the task's default)
+---
+specVersion: "4.0"
+kind: extension
+type: task
+metadata:
+  name: render-markdown-files
+task:
+  path: lib/tasks/renderMarkdownFiles.js
+  defaultBuildPhase: post-create-bundles # <- New default configuration
```
d3xter666, line 111: IMO, the `default` part is a bit missleading. Maybe, skipping it would make things clearer.
`defaultBuildPhase` seems like it implies changing the default phase for everything which is not the case here. Skipping `default` from the name makes it clear that this task will be executed at that phase

```suggestion
  buildPhase: post-create-bundles # <- New default configuration
```

PR #918, opened 2024-01-24
[INTERNAL] Setup v3 branch

flovogt:
JIRA: CPOUI5FOUNDATION-775


audit-ci.jsonc
```diff
@@ -0,0 +1,6 @@
+{
```
d3xter666, line 1: If this would be the default branch, then security audit would be handled directly by GitHub.

This configuration needs to be only inside the branch that would be scanned via `audit-ci`

d3xter666 (commented):
LGTM

RandomByte (approved):

PR #928, opened 2024-02-19
[BREAKING] Require Node.js 20.11.x/>=21.2.0 and npm >=10

RandomByte:
BREAKING CHANGE:
Support for older Node.js and npm releases has been dropped.
Only Node.js 20.11.x and >=21.2.0 as well as npm v10 or higher are supported.

JIRA: CPOUI5FOUNDATION-800

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -18,10 +18,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
 
-    - name: Use Node.js LTS 18.x
+    - name: Use Node.js LTS 20.x
       uses: actions/setup-node@v4.0.2
       with:
-        node-version: 18.x
+        node-version: 20.x
 
     - name: Install npm dependencies
       run: npm ci
--- a/scripts/isWorkspace.js
+++ b/scripts/isWorkspace.js
@@ -1,7 +1,7 @@
 import path from "node:path";
 import {fileURLToPath} from "node:url";
 
-const cliPath = path.relative(__dirname, fileURLToPath(import.meta.resolve("@ui5/cli/package.json")));
+const cliPath = path.relative(import.meta.dirname, fileURLToPath(import.meta.resolve("@ui5/cli/package.json")));
 
 // no workspace detected
 if (!cliPath.startsWith(path.join("..", ".."))) {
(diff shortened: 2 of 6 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -18,10 +18,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
 
-    - name: Use Node.js LTS 18.x
+    - name: Use Node.js LTS 20.x
       uses: actions/setup-node@v4.0.2
       with:
-        node-version: 18.x
+        node-version: 20.x
```
d3xter666, line 24: For the other repos we use `20.11.0`
Should that matter here?

scripts/isWorkspace.js
```diff
@@ -1,13 +1,7 @@
...
 import {fileURLToPath} from "node:url";
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
-// Using CommonsJS require.resolve as long as import.meta.resolve is experimental
-import {createRequire} from "node:module";
-const require = createRequire(import.meta.url);
-
-const cPath = require.resolve("@ui5/cli/package.json");
-const cliPath = path.relative(__dirname, path.dirname(cPath));
+const cliPath = path.relative(__dirname, fileURLToPath(import.meta.resolve("@ui5/cli/package.json")));
```
d3xter666, line 4: `__dirname` seems to be undefined here. Should it be `import.meta.dirname`?

```suggestion
const cliPath = path.relative(import.meta.dirname, fileURLToPath(import.meta.resolve("@ui5/cli/package.json")));
```

.github/workflows/deploy-docs.yml
RandomByte: In the other repositories we want to make sure that we always _test_ with the **minimal supported version** of Node.js, which is 20.11.0 and 21.2.0.

In this particular workflow, we don't test our modules but build the documentation. Using the latest LTS version seems more reasonable here, so that we can benefit from bug fixes and improvements.

scripts/isWorkspace.js
RandomByte: True! Thanks

matz3 (approved):

PR #931, opened 2024-02-26
[INTERNAL] Builder: generateFlexChangesBundle task is always enabled for applications

RandomByte:
This task was enabled since at least v2 (probably since v0 already):
https://github.com/SAP/ui5-builder/blob/v2.11.11/lib/types/application/ApplicationBuilder.js#L46

Not sure why we missed it in this list. Also fix formatting in the
markdown


flovogt (approved):
What about downporting this change also to v2 and v3 docu?

PR #933, opened 2024-03-06
[INTERNAL] Update @apidevtools/json-schema-ref-parser to latest version

flovogt:
In one of the former releaes, the json-schema-ref-parser did not export correctly the default export.
So we had to configure a fixed version before the bug was introduced.
This is fixed with https://github.com/APIDevTools/json-schema-ref-parser/releases/tag/v11.2.3, so we can use the latest version again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
 				"@ui5/server": "^3.1.5"
 			},
 			"devDependencies": {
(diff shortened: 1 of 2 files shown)


flovogt:
Generated the schema before and after this change. No differences were detected.

d3xter666 (approved):

PR #946, opened 2024-04-16
[RFC] 0016 Library Source Manifest

RandomByte:
Read: **[RFC 0016 Library Source Manifest](https://github.com/SAP/ui5-tooling/blob/rfc-library-source-manifest/rfcs/0016-library-source-manifest.md)**

JIRA: CPOUI5FOUNDATION-1155


flovogt:
just an idea: What about enhancing also the `init` command? to create a minimalistic `manifest.json`? 

e.g. `ui5 init --id my.fancy.app` or even with detection of namespace in the `Component.js` file?

PR #957, opened 2024-05-21
[FEATURE] Bundling: Remove option 'usePredefineCalls'

flovogt:
Up until UI5 Tooling v3, the bundle option "usePredefineCalls" defaults to "false" and has to be activated explicitly in a custom bundle configuration.
For default bundles, such as Component-preload or self-contained bundle (sap-ui-custom.js), there is no way to use the option apart from re-defining the whole bundle via custom bundle definition.

With UI5 Tooling v4 bundles are generated with the usage sap.ui.predefine calls instead of the former default function wrapper. This leads to smaller bundle sizes and less overhead at runtime.

As the option only affects the internal handling of bundling without affecting the actual users,
the option is removed completely, instead of just changing the default value of the option.

Implementation covered in https://github.com/SAP/ui5-builder/pull/1021.

JIRA: CPOUI5FOUNDATION-760


d3xter666 (approved):
LGTM

PR #963, opened 2024-06-19
[FEATURE] Reflect specVersion: 4.0 into documentation

d3xter666:
JIRA: CPOUI5FOUNDATION-835

https://github.com/SAP/ui5-tooling/issues/701

### DIFF
--- a/docs/pages/Configuration.md
+++ b/docs/pages/Configuration.md
@@ -16,7 +16,7 @@ See the list of [clients](https://github.com/redhat-developer/yaml-language-serv
 ## Example
 
 ```yaml
-specVersion: "3.2"
+specVersion: "4.0"
 type: application|library|theme-library|module
 metadata:
   name: some.project.name
@@ -36,28 +36,28 @@ The type defines the default path mappings and build tasks. See [UI5 Builder: Ty
     === "application"
 
         ```yaml
-        specVersion: "3.2"
+        specVersion: "4.0"
         type: application
         ```
 
     === "library"
 
         ```yaml
-        specVersion: "3.2"
+        specVersion: "4.0"
         type: library
         ```
 
     === "theme-library"
 
         ```yaml
-        specVersion: "3.2"
+        specVersion: "4.0"
         type: theme-library
         ```
 
     === "module"
 
         ```yaml
-        specVersion: "3.2"
+        specVersion: "4.0"
         type: module
         ```
 
@@ -270,7 +270,7 @@ You can find an overview of the available versions for each framework here:
 !!! example
     === "application"
         ```yaml
-        specVersion: "3.2"
+        specVersion: "4.0"
         type: application
         metadata:
           name: my.company.app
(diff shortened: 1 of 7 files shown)


docs/pages/Configuration.md
```diff
@@ -750,6 +751,26 @@ Version | UI5 CLI Release
 **1.0** | v1.0.0+
 **0.1** | v0.0.1+
 
+### Specification Version 4.0
+
+**Breaking changes:**
+
+- @ui5/builder: Remove bundle option [`usePredefineCalls`](#properties) 
+- @ui5/builder: ui5-builder: New option “async“ for [builder.bundles.bundleDefinition.section](#properties)
+- @ui5/builder: Task API "namespace" option got renamed to ["projectNamespace"](https://sap.github.io/ui5-tooling/v4/api/module-@ui5_builder_tasks_bundlers_generateStandaloneAppBundle.html)
```
RandomByte, line 760: This is supposed to only list changes in the Specification Version (i.e. the schema or functionality that explicitly checks the spec version).

API changes are usually independent from the Specification Version. For example this one affects projects of all Specification Versions and should therefore be mentioned in the UI5 Tooling 4.0 changelog but not here in the Specificatgion Version 4.0. Same for the rest of the chapter. See spec version 3.2 for an example.

RandomByte (changes requested):

docs/pages/Configuration.md
d3xter666: Thanks!

docs/pages/Configuration.md
```diff
@@ -750,6 +751,15 @@ Version | UI5 CLI Release
 **1.0** | v1.0.0+
 **0.1** | v0.0.1+
 
+### Specification Version 4.0
+
+**Breaking changes:**
+
+- Remove bundle option [`usePredefineCalls`](#properties)
```
RandomByte, line 758:
```suggestion
- Remove bundle option [`usePredefineCalls`](#properties). UI5 CLI v4.0.0 and above will always use predefine calls in bundles, making this option obsolete.
```

docs/pages/Configuration.md
```diff
@@ -750,6 +751,15 @@ Version | UI5 CLI Release
 **1.0** | v1.0.0+
 **0.1** | v0.0.1+
 
+### Specification Version 4.0
+
+**Breaking changes:**
+
+- Remove bundle option [`usePredefineCalls`](#properties) 
+- New option “async“ for [builder.bundles.bundleDefinition.section](#properties)
```
RandomByte, line 759: The async flag is not yet documented at the linked chapter. Should we do that in this PR or separately?

RandomByte (changes requested):

docs/pages/Configuration.md
RandomByte: Just to give some context on why it has been removed and what the behvavior looks like now.

docs/pages/Configuration.md
d3xter666: Thanks!
I'll add it with this PR

d3xter666: Done

RandomByte (approved):

PR #1001, opened 2024-08-09
[FIX] Updated node version requirements in docs

JensWallgren:
The docs had not been updated with the correct node version requirements after the v4 release.


matz3:
Thanks a lot 👍🏻 We've missed to update that page

PR #1056, opened 2025-05-12
ci: Use least privilege principle for permissions

flovogt:
Declares only the required permissions per workflow.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - main
+
+permissions:
+  contents: write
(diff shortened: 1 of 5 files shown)


matz3 (approved):

.github/workflows/deploy-docs.yml
```diff
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - main
+
+permissions:
+  contents: write
```
flovogt, line 8: @matz3 do we need this here?

flovogt: I guess we need it

PR #1097, opened 2025-09-10
[INTERNAL] Cleanup after Migration

flovogt:
JIRA: CPOUI5FOUNDATION-1058


flovogt:
The rename of the root level package will be done in CPOUI5FOUNDATION-975

PR #1119, opened 2025-09-25
docs: Adjust documents to reflect the new monorepo setup

d3xter666:
JIRA: CPOUI5FOUNDATION-975

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@
 
 ## 📝 Reporting Issues
 ### Seeking Help / Not a Bug
-If you need help setting something up, or if you have questions regarding UI5 CLI, please seek help on a community platform like [StackOverflow](http://stackoverflow.com/questions/tagged/ui5-tooling) or the [`#tooling`](https://openui5.slack.com/archives/C0A7QFN6B) channel of the [OpenUI5 Community Slack](https://ui5-slack-invite.cfapps.eu10.hana.ondemand.com).
+If you need help setting something up, or if you have questions regarding UI5 CLI, please seek help on a community platform like [StackOverflow](https://stackoverflow.com/questions/tagged/ui5-cli) or the [`#tooling`](https://openui5.slack.com/archives/C0A7QFN6B) channel of the [OpenUI5 Community Slack](https://ui5-slack-invite.cfapps.eu10.hana.ondemand.com).
 
 ### How to Report an Issue
 We collect issues for all UI5 CLI packages (*ui5-cli*, *ui5-server*, *ui5-project*, *ui5-builder*, *ui5-fs*, and *ui5-logger*) in [this](https://github.com/UI5/cli) monorepo. All packages are developed and maintained within this single mono repository.


CONTRIBUTING.md
```diff
@@ -6,15 +6,15 @@
...
-- [Pull Request Queue (cross-repo)](https://github.com/pulls?utf8=%E2%9C%93&q=is%3Apr+-label%3Adependencies+-label%3ARFC+is%3Aopen+no%3Aassignee+sort%3Aupdated-desc+repo%3AUI5%2Fcli+repo%3ASAP%2Fui5-cli+repo%3ASAP%2Fui5-logger+repo%3ASAP%2Fui5-project+repo%3ASAP%2Fui5-server+repo%3ASAP%2Fui5-builder+repo%3ASAP%2Fui5-fs+repo%3ASAP%2Fgrunt-openui5+repo%3ASAP%2Fless-openui5+repo%3ASAP%2Fconnect-openui5+repo%3ASAP%2Fkarma-openui5+repo%3ASAP%2Fopenui5-sample-app)
+- [All Open Pull Requests](https://github.com/UI5/cli/pulls?q=is%3Aopen+is%3Apr)
+- [Pull Request Queue](https://github.com/UI5/cli/pulls?q=is%3Apr+-label%3Adependencies+-label%3ARFC+is%3Aopen+no%3Aassignee+sort%3Aupdated-desc)
 
 ## 📝 Reporting Issues
 ### Seeking Help / Not a Bug
 If you need help setting something up, or if you have questions regarding UI5 CLI, please seek help on a community platform like [StackOverflow](http://stackoverflow.com/questions/tagged/ui5-tooling) or the [`#tooling`](https://openui5.slack.com/archives/C0A7QFN6B) channel of the [OpenUI5 Community Slack](https://ui5-slack-invite.cfapps.eu10.hana.ondemand.com).
```
maxreichmann: We maybe have to adjust the tags on Stack Overflow and Slack as well.

d3xter666: I'm not quite sure whether they have been transfered everywhere yet

maxreichmann (changes requested):
1. The legacy branch name here (which results in a dead link) needs to be replaced with "v3" here:
https://github.com/UI5/cli/blob/7fd601df832b18db52a516121d86eb50e81b8aae/docs/pages/extensibility/CustomTasks.md?plain=1#L205

2. There are some changelogs of the packages which contain outdated links e.g. here: https://github.com/UI5/cli/blob/7fd601df832b18db52a516121d86eb50e81b8aae/packages/builder/CHANGELOG.md?plain=1#L105 Due to our link forwarding, this might not be necessary :)

d3xter666:
> 1. The legacy branch name here (which results in a dead link) needs to be replaced with "v3" here:
>    https://github.com/UI5/cli/blob/7fd601df832b18db52a516121d86eb50e81b8aae/docs/pages/extensibility/CustomTasks.md?plain=1#L205
> 2. There are some changelogs of the packages which contain outdated links e.g. here: https://github.com/UI5/cli/blob/7fd601df832b18db52a516121d86eb50e81b8aae/packages/builder/CHANGELOG.md?plain=1#L105
>     Due to our link forwarding, this might not be necessary :)

I have updated the link in #1, but the second one is actually expected. There's a forwarding enabled in the old repo

maxreichmann:
> I have updated the link in #1, but the second one is actually expected. There's a forwarding enabled in the old repo

Although, we have this link forwarding, I think we should replace legacy links nevertheless. This might create confusion or even dead links if the forwarding is not working anymore at some point.
Perhaps, @flovogt knows more about this?

README.md
maxreichmann, line 23: Also, could you please adjust these links to use "v4" instead of "v3" while we're at it?

CONTRIBUTING.md
maxreichmann: Perhaps, @RandomByte does know more about this?

README.md
d3xter666: Done

CONTRIBUTING.md
RandomByte: SO seems to have been addressed. https://stackoverflow.com/questions/tagged/ui5-tooling will automatically search both tags and https://stackoverflow.com/questions/tagged/ui5-cli now holds the description. So we should link to https://stackoverflow.com/questions/tagged/ui5-cli

The slack channel doesn't need a renaming IMHO since it is also used for other "tooling" around UI5, not exclusively the CLI.

flovogt:
> > I have updated the link in #1, but the second one is actually expected. There's a forwarding enabled in the old repo
> 
> Although, we have this link forwarding, I think we should replace legacy links nevertheless. This might create confusion or even dead links if the forwarding is not working anymore at some point. Perhaps, @flovogt knows more about this?

Forwarding will work as long as we do not stop it. So we don't have to worry about this.

CONTRIBUTING.md
d3xter666: Done

PR #1123, opened 2025-09-25
refactor: Integrate @ui5/server into the monorepo

d3xter666:
JIRA: CPOUI5FOUNDATION-975


REUSE.toml
```diff
@@ -9,3 +9,15 @@ path = "**"
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2025 SAP SE or an SAP affiliate company and UI5 CLI contributors"
 SPDX-License-Identifier = "Apache-2.0"
+
+[[annotations]]
+path = "lib/middleware/serveIndex/**"
```
matz3: I wonder how/why this works. The file is at the root, and the paths start with `lib/`. But the code is located within `packages/fs/lib`, isn't it?

d3xter666: I suppose, it's because of the first rule which is very generic.. and still... it's with `MIT` and not `Apache`...

I have updated it with the correct path now

matz3 (approved):

PR #1126, opened 2025-09-26
refactor: Integrate @ui5/fs into the monorepo

d3xter666:
JIRA: CPOUI5FOUNDATION-975


d3xter666:
> It looks like the previously removed squash commit from logger is part of this PR. Is it possible to remove it?

After merging the logger, GitHub forced a rebase with the main branch and I did it via GH's button. Please take a look at the commit, it's clean

matz3:
The commit is not part of the `main` branch anymore, so we should also remove it from here.

d3xter666:
I needed to rebase on top of the main branch. Now should be better

matz3 (approved):

PR #1129, opened 2025-09-26
ci(coveralls): Workaround for 500 server errors

maxreichmann:


d3xter666:
This issue seems to be resolved now: https://github.com/coverallsapp/coverage-reporter/issues/180
Shall we postpone a bit to see whether the solution actually works before submitting this one?

maxreichmann:
> This issue seems to be resolved now: [coverallsapp/coverage-reporter#180](https://github.com/coverallsapp/coverage-reporter/issues/180) Shall we postpone a bit to see whether the solution actually works before submitting this one?

The issue still seems not to be resolved. In the original thread, another tracking issue has been created (https://github.com/lemurheavy/coveralls-public/issues/1824).

At the end, the overall problem is still the dependency to Coveralls which could break our ci pipelines in case of an error.

PR #1131, opened 2025-09-28
[FIX] Use Node's native https instead of spdy

### LINKED ISSUE #327: Use native http2 instead of spdy
## Expected Behavior
Native HTTP/2 protocol from Node.js is used.

## Current Behavior
SPDY protocol is used for HTTP/2 server.

## Details

This currently depends on the native HTTP/2 of express as it's used as the underlaying framework for our server: https://github.com/expressjs/express/pull/3730

However we might also switch to a different framework such as [fastify](https://github.com/fastify/fastify), which supports native HTTP/2 and out of the box provides an asynchronous middleware engine compatible with Express and Restify middlewares (https://github.com/fastify/fastify/blob/master/docs/Middlewares.md) so we wouldn't need to re-write our existing middlewares.

henriquemattos:
Using native `https` from NodeJS instead of SPDY, which is not supported by Node >= 24.

According to [NodeJS Releases](https://nodejs.org/en/about/previous-releases), “v22 is about to enter Maintenance and v24 is about to become Active”. Unfortunately, this is a breaking change for UI5 CLI `serve` when using `--h2`, since we use `spdy`. According to the code comments, we use `spdy` because of Node v8.4.0 😬.

Tested the upgrade using Firefox Developer Edition (latest) on macOS 26 and NodeJS v24.9.0 with:
```shell
ui5 serve --verbose --port 1987 -o index.html --h2 --key ./infrastructure/certs/localhost.key --cert ./infrastructure/certs/localhost.crt --accept-remote-connections
```

Fixes: #327

--- 

**Thank you for your contribution!** 🙌

To get it merged faster, kindly review the checklist below:

## Pull Request Checklist
- [X] Reviewed the [Contributing Guidelines](https://github.com/UI5/cli/blob/main/CONTRIBUTING.md#-contributing-code)
    + Especially the [How to Contribute](https://github.com/UI5/cli/blob/main/CONTRIBUTING.md#how-to-contribute) section 
- [X] [No merge commits](https://github.com/UI5/cli/blob/main/docs/Guidelines.md#no-merge-commits)
- [X] [Correct commit message style](https://github.com/UI5/cli/blob/main/docs/Guidelines.md#commit-message-style)

### DIFF
--- a/packages/server/lib/server.js
+++ b/packages/server/lib/server.js
@@ -1,4 +1,5 @@
 import express from "express";
+import https from "https";
 import portscanner from "portscanner";
 import MiddlewareManager from "./middleware/MiddlewareManager.js";
 import {createReaderCollection} from "@ui5/fs/resourceFactory";
@@ -87,10 +88,7 @@ function _listen(app, port, changePortIfInUse, acceptRemoteConnections) {
  * @private
  */
 async function _addSsl({app, key, cert}) {
-	// Using spdy as http2 server as the native http2 implementation
-	// from Node v8.4.0 doesn't seem to work with express
-	const {default: spdy} = await import("spdy");
-	return spdy.createServer({cert, key}, app);
+        return https.createServer({key: key, cert: cert}, app);
 }
 
 
@@ -180,12 +178,6 @@ export async function serve(graph, {
 	await middlewareManager.applyMiddleware(app);
 
 	if (h2) {
-		const nodeVersion = parseInt(process.versions.node.split(".")[0], 10);
-		if (nodeVersion >= 24) {
-			log.error("ERROR: With Node v24, usage of HTTP/2 is no longer supported. Please check https://github.com/UI5/cli/issues/327 for updates.");
-			process.exit(1);
-		}
-
 		app = await _addSsl({app, key, cert});
 	}
 


flovogt:
@henriquemattos Thanks a lot for your contribution.

Note: This PR is a breaking change and would also drop HTTP/2 support but ensure HTTPS is working again.

henriquemattos:
I see your point, @flovogt . Thank you for the feedback. I see HTTP/2 in development environment as a “nice-to-have” feature, while SSL is, at least for Firefox, a “must-have”. The default settings of the browser uses https even for localhost. Besides, changing the protocol on the frontend would require a CORS, Proxy or change to the backend too.

Based on your comment, I don’t understand what are the next steps?!

kind regards,
Henrique

RandomByte:
Thanks again for your contribution. We are currently discussing several options in the team on how to proceed with HTTPS support in UI5 CLI.

Since this is a breaking change, we are thinking of actually going one step further and replacing express with connect. This would allow us to use the native [Node.js HTTP/2 server](https://nodejs.org/api/http2.html) directly and also solve this issue.

We plan major changes to the UI5 Server in UI5 CLI v5 and never made great use of the APIs express provides over connect.

>  while SSL is, at least for Firefox, a “must-have”. The default settings of the browser uses https even for localhost.

@henriquemattos I'm not able to reproduce this with Firefox 1.145. Could you share some details or documentation on this? 

The current documentation on [Secure Context](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#potentially_trustworthy_origins) classifies localhost as "potentially trustworthy", even over HTTP. In my test, I could use for example the [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) on `http://localhost`

henriquemattos:
Thanks for the update, @RandomByte . 

As for Firefox, I use the Developer Edition and somehow it came with the HTTPS-Only mode enabled. This was becoming a hassle not only for my localhost, but acccessing my router's IP address previously.

https://support.mozilla.org/en-US/kb/https-only-prefs

RandomByte:
I see, thanks. We hope we can provide a solution for current Node.js versions in UI5 CLI v5, one way or another.

henriquemattos:
In the meantime, do you think would be possible to apply my solution via middleware? I mean, every time I update Node modules, I have to reimplement the workaround locally (as I’m already developing in Node v24 for the next quarter).

RandomByte:
> In the meantime, do you think would be possible to apply my solution via middleware? I mean, every time I update Node modules, I have to reimplement the workaround locally (as I’m already developing in Node v24 for the next quarter).

Sorry for not replying to this. Since this is a breaking change we can't implement it in UI5 CLI v4. We are currently working on v5 where we still hope to switch over to the native Node.js HTTP/2 server as mentioned before.

Reading your description regarding Firefox again, I'm wondering whether you have an HSTS entry for localhost that forces Firefox to use HTTPS. You might want to try to "Clear cookies and site data" for `localhost:8080` in your Firefox. Similarly, you can try whether you are also forced to HTTPS in a private tab opening `localhost:8080`.

henriquemattos:
@RandomByte no worries, but thank you for getting back. In the meantime, @flovogt had already updated me on Slack about v5.

And yes, Firefox eventually built a “HTTPS-only” toggle and it was switched on (not sure if by default or by me, as it has been a while).

If it helps a decision on your end, I would be fine with a flag `--ssl` as a subset of H2, starting `https` instead of `spdy`. I’m running development workload on UI5 CLI (and the `serve` shows a message it’s not intended for production).

Thank you and best regards,
Henrique

PR #1147, opened 2025-10-09
docs: Enable documentation ci

d3xter666:
This PR enables the CI integration and build of the VitePress docs

JIRA: CPOUI5FOUNDATION-899

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Generate JSDoc documentation
       working-directory: packages/documentation
       run: |
-        npm ci
+        npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
         npm run jsdoc-generate
 
     - name: Generate merged JSON schema


d3xter666:
A deployed version of the documentation with this pipeline can be seen here: https://d3xter666.github.io/test-mono/v5/

maxreichmann:
Looks very good. The only thing I could find is the home page linking to itself at the bottom as "Next Page":
Screenshot 2025-10-09 at 12 52 00

maxreichmann (dismissed):

REUSE.toml
```diff
@@ -26,4 +26,10 @@ SPDX-License-Identifier = "Apache-2.0"
...
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2025 SAP SE or an SAP affiliate company and OpenUI5 contributors"
-SPDX-License-Identifier = "Apache-2.0"
\ No newline at end of file
+SPDX-License-Identifier = "Apache-2.0"
+
+[[annotations]]
+path = "packages/documentation/.vitepress/**"
+precedence = "aggregate"
+SPDX-FileCopyrightText = ["2019-present, Konrad Kost, Yuxi (Evan) and VitePress contributors"]
```
RandomByte: No need to mention Holger Schäfer here?

d3xter666: I've added him. Thanks!

d3xter666:
The result of the deployment can be seen here: https://d3xter666.github.io/ui5-tooling/v5/

matz3:
> The result of the deployment can be seen here: https://d3xter666.github.io/ui5-tooling/v5/

The version dropdown doesn't seem to work properly, but it could be that this is just an issue with your deployment, as it uses a different repo name (`ui5-tooling` instead of `cli`). This also seems to be the reason why the image is not loaded on the index page.

d3xter666:
> > The result of the deployment can be seen here: https://d3xter666.github.io/ui5-tooling/v5/
> 
> The version dropdown doesn't seem to work properly, but it could be that this is just an issue with your deployment, as it uses a different repo name (`ui5-tooling` instead of `cli`). This also seems to be the reason why the image is not loaded on the index page.

Now all the links should be resolved relatively, based on the current deployment!
There are 2 deployments: `v5` and `next`:
- https://d3xter666.github.io/ui5-tooling/v5/
- https://d3xter666.github.io/ui5-tooling/next/

**Note**: v4 is not available in this repo

We must deploy twice with modified config as vitepress uses this information to build links and some links might get broken if we just use symlink for next as it used to be with the previous docs deployment.

Also the 404 redirect issues have been addressed. Now the behaviour is:
- if the version exists, but the page does not- redirect to the 404 page
- if the version doesn't exist, redirect to the `stable` version

.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -0,0 +1,99 @@
...
+                  npm run schema-workspace-generate
+            - name: Checkout gh-pages
+              uses: actions/checkout@v4
+              with:
+                  ref: gh-pages
+                  path: gh-pages
+            - name: Copy the additional resources to gh-pages
+              run: |
+                  rm -rf ./gh-pages/schema
+                  cp -R ./site/schema ./gh-pages/
```
matz3: This workflow now also updates the central `schema` folder and the custom 404 page. The current v4 workflow does the same, so they might overwrite the files when deploying from main / v4: https://github.com/UI5/cli/blob/v4/.github/workflows/deploy-docs.yml

.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -0,0 +1,99 @@
...
+                  rm -rf ./gh-pages/${DOC_ALIAS}
+
+                  # Main version route
+                  cp -R ./packages/documentation/dist ./gh-pages/${DOC_VERSION}/
+                  cp -R ./site/api ./gh-pages/${DOC_VERSION}/api
+
+                  # Alias route. E.g., next, latest, stable, etc.
+                  # For vitepress must be a copy with different config, not a symlink
+                  cp -R ./packages/documentation/dist-${DOC_ALIAS} ./gh-pages/${DOC_ALIAS}/
+                  cp -R ./site/api ./gh-pages/${DOC_ALIAS}/api
```
matz3: The API reference is just linked from vitepress but not included in the build. Isn't it possible to at least keep using a symlink here?

d3xter666: Please, take a look here: https://github.com/UI5/cli/pull/1147/files/2b5f2a0bdb4f8ddb5925a8a4bf2750bc4e19f7a2#diff-017c3adbedb4d6453fdcfcee89d833b8a202fa21ecbc048ff5133b52deeb8103R56
It's the API ref build.

And this is the CI pipline build for it:
- v5: https://d3xter666.github.io/ui5-tooling/v5/api/index.html
- next: https://d3xter666.github.io/ui5-tooling/next/api/index.html

matz3: So can't we create a symlink between `next/api` and `v5/api` instead of copying the files?

d3xter666: Yep, thanks!
Now there's a symlink in the next branch for the API folder and it seems to be wokring correctly: https://d3xter666.github.io/ui5-tooling/next/api/index.html

By "and `v5/api`" do you mean that we should keep the `API` folder somewhere and reference it within `v5` and `next` folders?

Current `mike` implementation actually has a copy for every version:
i.e. https://github.com/UI5/cli/tree/gh-pages/v3 -> https://github.com/UI5/cli/tree/gh-pages/v3/api

matz3: No, I just meant what you've applied: Adding the symlink instead of copying the folder. LGTM.

.github/workflows/deploy-vitepress-docs.yaml
d3xter666: I have updated the workflow, so that it now does not deploy schema when building documentation for v5.

Further solution on how schema will be generated and deployed is described in CPOUI5FOUNDATION-1162

d3xter666:
The latest deployments with the JSDoc integration into the `documentation` package and moved CLI generation script can be seen here:
- https://d3xter666.github.io/ui5-tooling/next/
- https://d3xter666.github.io/ui5-tooling/v5/

.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -0,0 +1,100 @@
...
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
+              with:
+                  fetch-depth: 0
+            - name: Setup Node
+              uses: actions/setup-node@v4
+              with:
+                  node-version: 20
```
matz3: https://github.com/UI5/cli/blob/2074a75f835795e63e8e1b1ecc21f45c52760235/package.json#L22
https://github.com/UI5/cli/blob/2074a75f835795e63e8e1b1ecc21f45c52760235/packages/documentation/package.json#L21

.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -0,0 +1,100 @@
...
+                  # Main version route
+                  cp -R ./packages/documentation/dist ./gh-pages/${DOC_VERSION}/
+
+                  # Alias route. E.g., next, latest, stable, etc.
+                  # For vitepress must be a copy with different config, not a symlink
+                  cp -R ./packages/documentation/dist-${DOC_ALIAS} ./gh-pages/${DOC_ALIAS}/
+                  # Symlink the api docs to avoid duplication
+                  ln -s ../${DOC_VERSION}/api ./gh-pages/${DOC_ALIAS}/api
+
+                  cp ./packages/documentation/scripts/resources/custom404.html ./gh-pages/404.html
```
matz3: This means we have to disable the update of 404.html from v4 branch, right? Otherwise a v4 deployment would overwrite it again.

d3xter666: I will disable it for now here as it will be easier to do the whole disable/enable step at once for v4 -> v5.

.github/workflows/deploy-vitepress-docs.yaml
d3xter666: Thanks! I have aligned not just this prop, but all of the packages with the `github-ci.yml`

.github/workflows/github-ci.yml
```diff
@@ -40,10 +40,14 @@ jobs:
       run: npm run coverage
 
     - name: Generate JSDoc documentation
-      run: cd packages/documentation && npm ci && npm run jsdoc-generate
+      working-directory: packages/documentation
+      run: |
+        npm ci
```
matz3: As you've made changes to the deploy workflow, I'd suggest also adding the same checks to the CI workflow, and adding `--engine-strict` here

.github/workflows/deploy-vitepress-docs.yaml
matz3: So the changes to the file are not relevant until then?

d3xter666: The issue with the 404 is that with the old implementation we were relying on the following pattern: `v{0-9}|next|stable` withtout actually knowing whether certain version exists. So, if you modify the URL, so that it points to an non existing version, it tries to fix it by redirecting to the root page. As the version does not exist, it tries to redirect to the stable version. And this redirection to the stable version is broken. 

That's why this solution is:
- if the version exists, but the page does not- redirect to the 404 page
- if the version doesn't exist, redirect to the stable version

Basically, without an URL modification, the issue cannot be reproduced.
This means that if someone tries to guess certain version that does not exist, it will currently also break the production. The solution does not relate to the neew deployment, it was just exposed by the demo I have prepared. That specific deployment was missing the v4, but there was a v4 link.

.github/workflows/github-ci.yml
d3xter666: Thanks!
Done

PR #1148, opened 2025-10-10
ci: Integrate "project" into monorepo setup

maxreichmann:
JIRA: CPOUI5FOUNDATION-1139

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 * text=auto eol=lf
 
-# Needed for packages/project
+# Needed for packages/project, see https://github.com/SAP/ui5-project/pull/679
 packages/project/build/** linguist-generated=false
 packages/project/lib/build/** linguist-generated=false
 packages/project/lib/** linguist-vendored=false


packages/project/package.json
```diff
@@ -151,19 +146,14 @@
 		}
 	},
```
maxreichmann, line 147: @UI5/ui5-team-cor-fnd These settings stay untouched?

d3xter666: IMO, yes! These are on package level i.e. when installing the @ui5/project and checking its dependencies.

RandomByte: Yes, they should remain as-is 👍

.gitattributes
```diff
@@ -1 +1,7 @@
 * text=auto eol=lf
+
+# Needed for packages/project
```
RandomByte:
```suggestion
# Needed for packages/project, see https://github.com/SAP/ui5-project/pull/679
```

I'm actually not sure whether this is still needed, but it probably won't hurt either. So let's keep it but add a reference so we understand why it's there.

maxreichmann:
*Rebased*

PR #1157, opened 2025-10-17
fix: Provide logo for README

d3xter666:
README.md files reference the root of the project for the UI5 Logo. Now it's provided generally for all the repositories


maxreichmann:
~~It looks like root README is working but the packages' READMEs aren't~~

Edit: The packages' READMEs use an absolute ("https://...") link using the main branch instead of a relative one. That's why it's "not working" in this PR. So nevermind my earlier comment :D

LGTM :)

PR #1168, opened 2025-10-22
ci: Add test matrix (Node 20, 22, 24 on Linux / macOS / Windows)

matz3:
Also update checkout action to latest version.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,36 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    name: Unit and Integration
+    strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Setup Node.js
+      uses: actions/setup-node@v6
+      with:
+        node-version: ${{ matrix.version }}
+
+    - name: Install dependencies
+      run: npm ci
+
+    - name: Run unit tests
+      run: npm run unit
```
d3xter666, line 36: Is there a reason for running just the unit tests, but not `npm run test`?

matz3: We already run all steps from "npm test" in the general workflow: https://github.com/UI5/cli/blob/main/.github/workflows/github-ci.yml

The matrix is intended to cover potential issues with the unit tests on different platforms / Node versions.

Are there some additional checks that you think should be run within the matrix?

PR #1172, opened 2025-10-27
ci: Fail-fast "npm ci" if deps require unsupported node version

maxreichmann:
`+` ci(package.json): Update node version to current LTS (22) + latest (24)
`+` ci(github-ci.yml): Update node version to current LTS (22)
`+` ci(test.yml): Remove node20 for test matrix

---

When upgrading a dependency to a newer version, it might require a different node version than before which could be unsupported by the repo. Unfortunately, this could be easily overseen.

This PR adds the flag `--engine-strict` to the dependency installation step during the GH Actions job. Incase such a version requirement change, this flag will now fail-fast the job which should attract attention.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false # Do not stop other jobs if one fails
       matrix:
-        version: [20, 22, 24]
+        version: [22, 24]
         os: [ubuntu-24.04, windows-2025, macos-15]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


maxreichmann:
@UI5/ui5-team-cor-fnd 
Currently, the tests in the matrix with Node20 fail because `Execa` detects the unsupported version. If we want to bump our node version requirements, we have to adjust the versions in the matrix too (even though we aren't even using `--engine-strict` here). Or do you have other ideas / opinions?

RandomByte (approved):

PR #1176, opened 2025-10-31
ci(release-please): Release please single PR

d3xter666:
JIRA: CPOUI5FOUNDATION-1142
JIRA: CPOUI5FOUNDATION-1164

This change modifies several things, so the whole workflow can be complete:
- Creates a Release Please process
    - Adds release please configuration and pipeline
    - Releases the all of the package at once with a single PR
    - Generates an `npm-shrinkwrap.json` for the CLI package before being released (out of monorepo's root `package-lock.json`)
- Introduce a new package `shrinkwrap-extractor` to deal with `@ui5/cli`'s `npm-shrinkwrap.json` generation in a monorepository environment
- Clear separation between "public" and "internal" packages
- Adjusts pipelines to deal with the new structure:
    - github-ci.yml
    - reuse-compliance.yml
- Adjust `deploy-vitepress-docs.yaml` so that it will generate API documentation out of published NPM packages and not from the packages within the monorepo
- Align `package.json` scripts, so that they can be easily (re)used within the monorepo structure.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -32,11 +32,8 @@ jobs:
     needs: release-please
     if: needs.release-please.outputs.releases_created == 'true'
     strategy:
-      # Publish packages in dependency order: logger → fs → builder → server → project
-      # CLI is handled separately to update shrinkwrap
-      # Order of packages in the matrix does matter and is important!
-      # release-please updates the package.json dependencies with ones that do not exist yet on NPM
-      # Publish in the correct order is important, so that dependencies are available and resolved properly!
+      # Sequential publishing ensures dependencies exist on NPM before dependents are published
+      # Order: logger → fs → builder → server → project (CLI handled separately for shrinkwrap generation)
       max-parallel: 1
       matrix:
         package: [logger, fs, builder, server, project]


d3xter666:
Please find a sample result of this flow in here: https://github.com/d3xter666/ui5-tooling/pull/139

**Note:** Please ignore the title, it has been updated later in the config

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,51 @@
...
+        package: [logger, fs, builder, server, project, cli]
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Node.js LTS
+        uses: actions/setup-node@v5
+        with:
+         node-version: 24.x
+
+      - name: Install and publish ${{ matrix.package }}
```
RandomByte: Special handling for the CLI package is required, where the shrinkwrap needs to be updated/created after all dependencies have been released. Should this be done in a separate PR?

matz3:
Please see my comment here: https://github.com/d3xter666/ui5-tooling/pull/139#issuecomment-3473211965

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Node.js LTS
+        uses: actions/setup-node@v5
+        with:
+         node-version: 24.x
+
+      - name: Run Release Please to update PRs and create releases
+        uses: googleapis/release-please-action@v4
+        id: release
+        with:
+          token: ${{ secrets.RELEASE_PLEASE_TOKEN }}
```
matz3: Isn't the default token from GitHub sufficient here?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,84 @@
...
+          node-version: 24.x
+
+      - name: Install and publish ${{ matrix.package }}
+        run: |
+          cd packages/${{ matrix.package }}
+          echo "🚀 Publishing @ui5/${{ matrix.package }}"
+          
+          # TODO: Uncomment when ready to publish
+          # npm ci
+          # npm publish --access public
```
matz3, line 57: OIDC for trusted publishing seems to be missing, see https://docs.npmjs.com/trusted-publishers and the workflow of linter / mcp-server / cli-extensions

.release-please-manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+	"packages/logger": "4.0.2",
```
matz3, line 2: I'd say we should adjust the versions accordingly, before merging this PR. Otherwise release please will create release PRs for v4, which can be very confusing.

d3xter666: Do you mean, to bump all package versions to 5.0.0?

.github/workflows/release-please.yml
d3xter666: Thanks! It's a copy/paste error as I'm trying to verify the build in another repo. I'll fix that!

.github/workflows/release-please.yml
d3xter666: Thanks!
I think we had something related to the shrinkwrap itself, but I can't find the discussion/BLI!

I have updated the action so that it:
- Runs the publishing of all the packages in a sequence. This ensures that during `npm ci` for publishing for the next package in the chain, it will have the previous package already published on NPM.
- For the CLI, it has a special step where all the packages are available already on NPM and it's been updated accordingly.

As per my research this is the approach many monorepos are wokring. Lerna, for example

.release-please-manifest.json
matz3: Yes, or rather to a pre-release version.

.github/workflows/release-please.yml
d3xter666: Done

.github/workflows/release-please.yml
d3xter666: Done

.release-please-manifest.json
d3xter666: Done: https://github.com/UI5/cli/pull/1176/commits/d67b4a515fa73495568885260f930d7bf44e9595

release-please-config.json
```diff
@@ -0,0 +1,94 @@
+{
+	"$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+	"include-component-in-tag": true,
```
matz3: Do we need this feature if we always release everything under the same version?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          node-version: 24.x
+
+      
+      - name: Prepare production dependency tree
+        working-directory: packages/cli
+        run: |
+          echo "📦 Installing production dependencies for CLI"
+          npm ci --omit=dev
+      - name: Generate npm-shrinkwrap.json (prod only)
```
matz3: Does this guarantee that the exact same transitive dependency versions from `package-lock.json` are used in the `packages/cli/npm-shrinkwrap.json`?

release-please-config.json
d3xter666: Yes, it's actually this description:
https://github.com/d3xter666/ui5-tooling/pull/140
vs this description:
https://github.com/d3xter666/ui5-tooling/pull/139

See how in the older PR does not include the component names and it's quite confusing what we actually release

.github/workflows/release-please.yml
d3xter666: Thanks for challenging this!
Now, the answer is Yes!

I have reworked it, so that it:
1. runs `npm ci` for the monorepo -> Creates a package-lock.json
2. then isolates the cli out of the monorepo (`npm shrinkwrap` does not work properly for workspaces) 
3. prunes unneeded dependencies from it
4. creates `npm shrinkwrap` for the isolated cli package
5. copies the generated `npm-shrinkwrap.json` back to the `packages/cli`
6. Deploys `cli` to npm

release-please-config.json
matz3: Yes, that is very confusing. However, wouldn't this flag cause the creation of git tags for each package in each release? E.g. `cli-v5.0.0`, `builder-v5.0.0`, ...? As we would always release all packages with the same version, this wouldn't be necessary, as a tag named `v5.0.0` would point to this version for all packages.

I'm also a bit confused by the flag description: `When tagging a release, include the component name as part of the tag. Defaults to `true`.`. This doesn't seem to mention the differences in the release PR summary.

.github/workflows/release-please.yml
matz3: I have tested this locally and compared the two lockfiles to see differences in package versions. Obviously they won't have the same content, but it helps to identify whether transitive dependencies are using the exact same version as in the existing `package-lock.json`. This is what I saw:

1. The devDependencies are still included and would get installed by consumers (see https://github.com/UI5/linter/blob/01500503f8063ea9e639939ca6c515b332305580/.github/workflows/release-please.yml#L40-L54 for our current solution to this problem).
2. I found at least one example where a transitive dependency uses a different (newer) version than defined in the `package-lock.json`. For https://www.npmjs.com/package/ci-info the shrinkwrap was using `4.3.1`, while currently our lockfile uses `4.3.0` . See: https://github.com/UI5/cli/blob/e0b185d886c21b52a74eab9bdf7a8e0bd5a17408/package-lock.json#L5829

I'm very unhappy about the missing support from npm for such a scenario, but I don't see that we should use such a setup as it defeats the purpose of using a lockfile in the project. We will end up shipping a release with dependency version combinations that were never tested before. If there happens to be no solution to this problem, I would rather prefer to stop using a shrinkwrap, as it would at least allow consumers to upgrade the transitive packages on their end if issues occur. But not using a shrinkwrap for CLI packages is IMHO not a good idea at all and I would really like to continue using it.

RandomByte:
For the shrinkwrap tool, I was wondering how we best integrate it in the monorepo. Maybe similar to the documentation as a dedicated package? Not sure whether `lib/` is the best choice, since it's not really that important for the UI5/cli project.

.github/workflows/release-please.yml
d3xter666: Now the shrinkwrap generation is handled with an additional script.
In this situation it seems that extraction of `npm-shrinkwarp.json` from `package-lock.json` is the most convenient and stable solution.

.github/workflows/release-please.yml
d3xter666: The discussion is consolidated here: https://github.com/UI5/cli/pull/1176#discussion_r2498844352

lib/shrinkwrap-extractor/eslint.config.js
matz3, line 1: Why not re-using the `eslint.common.config.js`?

lib/shrinkwrap-extractor/.editorconfig
matz3, line 1: We already have an editorconfig, gitattributes and npmrc at root level. I don't see a need for those 3 files.

lib/shrinkwrap-extractor/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "shrinkwrap-extractor",
+  "version": "1.0.0",
```
matz3: private flag should be set as we currently don't intend to publish this package, right?

d3xter666: Done

lib/shrinkwrap-extractor/eslint.config.js
d3xter666: done

lib/shrinkwrap-extractor/package.json
d3xter666: Done

lib/shrinkwrap-extractor/.editorconfig
d3xter666: Done

matz3 (commented):
My points have been addressed, but the comment by @RandomByte might still be open?

d3xter666:
> For the shrinkwrap tool, I was wondering how we best integrate it in the monorepo. Maybe similar to the documentation as a dedicated package? Not sure whether `lib/` is the best choice, since it's not really that important for the UI5/cli project.

I don't have strong oppinion on how to handle this one. Here's the reasoning why it felt to me the right way to go:
- The documentation package is "self contained" (except maybe for the `CLI.md` generation) and felt naturally to do it in a separarte package
- The shrinkwrap tool has the `@ui5/cli` package "dependency" and the `package-lock.json` from the monorepo. So, I was wondering whether it's best to be part of any of those "packages". And as we want to ship the `@ui5/cli` as clean as possible, the natural choice was to go in the `lib/` folder of the monorepo.

I'm curious to hear your oppinion on that.
If we decide now we'll pave the way for constant future structure of the monorepo

RandomByte:
My understanding is that the shrinkwrap tool does not have a "dependency" to `@ui5/cli`. In fact, you can use it to extract the dependencies for any package from any given package-lock.json. We could even publish it as a dedicated tool, since other mono-repositories face similar problems.

Personally, I find it unconventional to put a helper tool into the root `lib/` directory. In the UI5 linter project for example, `lib/` contains the transpiled code of the linter: https://www.npmjs.com/package/@ui5/linter?activeTab=code

I would vote for putting it into a package similar to documentation.

d3xter666:
> * The shrinkwrap tool has the `@ui5/cli` package "dependency" and the `package-lock.json` from the monorepo. So, I was wondering whether it's best to be part of any of those "packages". And as we want to ship the `@ui5/cli` as clean as possible, the natural choice was to go in the `lib/` folder of the monorepo.

Now, the `shrinkwrap-extractor` is a package within the `packages` folder. All the pipelines are adjusted, so that tests and integrations of the `shrinkwrap-extractor` are also executed

.github/workflows/github-ci.yml
```diff
@@ -51,3 +51,9 @@ jobs:
     - name: Generate CLI documentation
       working-directory: packages/documentation
       run: npm run generate-cli-doc
+
+    - name: Check shrinkwrap integrity
+      working-directory: packages/shrinkwrap-extractor
+      run: |
+        npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
+        npm run unit # The lint is done generally for the whole repo, so only unit tests are needed here
```
matz3: > The lint is done generally for the whole repo, so only unit tests are needed here

How is that? The package is not part of the workspace, so `npm run lint --workspaces` won't lint this projects. It is the same reason why unit tests are not executed.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Node.js LTS
+        uses: actions/setup-node@v5
+        with:
+          node-version: 24.x
+
+      - name: Generate npm-shrinkwrap.json
+        run: |
+          set -e
+          # Install dependencies
+          npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
```
matz3: Do we really need `--engine-strict` everywhere? Here, we are using Node 24, so the fundamental idea of this check, which is to prevent us from using prod dependencies that do not work on our minimum required Node.js version, will not apply.

packages/shrinkwrap-extractor/test/lib/convertToShrinkwrap.js
```diff
@@ -0,0 +1,216 @@
...
+	const expectedShrinkwrap = await readJson(expectedShrinkwrapPath);
+
+	// Write generated shrinkwrap to tmp dir for debugging purposes
+	await writeFile(generatedShrinkwrapPath, JSON.stringify(generatedShrinkwrap, null, "\t"), "utf-8");
+
+	assert.deepEqual(generatedShrinkwrap.packages, expectedShrinkwrap.packages,
+		"Generated shrinkwrap packages should match expected");
+});
+
+// // Error handling tests
```
matz3: Can we at least add the reason why those tests are not active?

.github/workflows/release-please.yml
d3xter666: Done

.github/workflows/github-ci.yml
d3xter666: Thanks!
It was a leftover from the previous destination `libs/`. Now it's fixed to run the whole spectrum

packages/shrinkwrap-extractor/test/lib/convertToShrinkwrap.js
d3xter666: It used to be a leftover from the implementation from the other, private repo. 
It was just that the signature of the function that changed.
I did updates and enabled the tests as they truely make sense- error handling

internal/documentation/package-lock.json
matz3: The package is now part of the workspace, so the package-lock should be removed.

.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -40,18 +40,23 @@ jobs:
...
-              run: npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
-            - name: Install documentation dependencies
-              working-directory: packages/documentation
-              run: npm ci --engine-strict # --engine-strict is used to fail-fast if deps require node versions unsupported by the repo
+            - name: Copy documentation package to isolated directory
+              run: |
+                  # Create isolated directory for documentation
+                  # So, that API.json will be build from the NPM packages,
+                  # and not from workspace packages
+                  mkdir -p /tmp/isolated-docs
```
RandomByte: Maybe better to use runner.temp to get a temporary directory as documented here: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#example-usage-of-the-runner-context

d3xter666: Thanks!

d3xter666:
I did some "local" tests in order to see whether the release process works correctly:
- With the help of [`verdaccio`](https://verdaccio.org/) I was able to replicate the the `.github/workflows/release-please.yml` workflow and to correctly publish the correct versions of the packages
- Release please [config](https://github.com/UI5/cli/pull/1176/files#diff-c55d4dcb68c348b2c06d263819702fbce72eeeb35b662956d02049a5a18fcf2bR6) helps with the correct release of the versions. You can take a look at an example PR with this configuration here: https://github.com/d3xter666/ui5-tooling/pull/140. All of the versions and interdepednecnies in `package.json` are correctly set as well as `package-lock.json` and `.release-please-manifest.json`
- Documentation's API is correctly build from NPM published resources.
    - The deployed docs: 
        - v5: https://d3xter666.github.io/ui5-tooling/v5/api/index.html 
        - next: https://d3xter666.github.io/ui5-tooling/next/api/index.html
        - Screenshot 2025-11-18 at 13 03 07

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,88 @@
...
+          token: "${{ secrets.GITHUB_TOKEN }}"
+
+  publish-packages:
+    runs-on: ubuntu-24.04
+    needs: release-please
+    if: needs.release-please.outputs.releases_created == 'true'
+    strategy:
+      # Publish packages in dependency order: logger → fs → builder → server → project
+      # CLI is handled separately to update shrinkwrap
+      # Order of packages in the matrix does matter and is important!
```
matz3: this line ...

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,88 @@
...
+      # release-please updates the package.json dependencies with ones that do not exist yet on NPM
+      # Publish in the correct order is important, so that dependencies are available and resolved properly!
```
matz3: ... and this line are basically saying the same, isn't it?

matz3 (commented):
LGTM otherwise

.github/workflows/release-please.yml
d3xter666: Done

.github/workflows/release-please.yml
d3xter666: I have reworded it. Thanks!

d3xter666: I have reworded it. Thanks!

PR #1179, opened 2025-11-03
ci: Add --engine-strict to remaining job configs

maxreichmann:


RandomByte:
IMHO it's enough to have the flag in a single workflow, ideally the one with the lowest Node.js version

maxreichmann:
Not planned atm.
Closing.

PR #1184, opened 2025-11-05
docs: Setup JSDoc build for monorepo

maxreichmann:
JIRA: CPOUI5FOUNDATION-1149 <br/>
image

---
#### Decisions taken
All JSDoc-related artifacts have been consolidated and moved into a combined directory (`/packages/documentation/jsdoc`).

The page can be built by running `npm run jsdoc` in `/packages/documentation/`).

The new build output path is`/packages/documentation/dist/api`.

The two layout/CSS files have been moved from root level to a new directory called "templates" under `/packages/documentation/jsdoc`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
 		"schema-generate": "node ./scripts/buildSchema.js ui5",
 		"schema-workspace-generate": "node ./scripts/buildSchema.js ui5-workspace",
 		"generate-cli-doc": "node ./scripts/generateCliDoc.js",
-		"depcheck": "depcheck --ignores @ui5/builder,@ui5/cli,@ui5/fs,@ui5/logger,@ui5/project,@ui5/server,docdash,local-web-server,@commitlint/config-conventional,husky && npm run depcheck --workspaces",
+		"depcheck": "depcheck --ignores @ui5/builder,@ui5/cli,@ui5/fs,@ui5/logger,@ui5/project,@ui5/server,local-web-server,@commitlint/config-conventional,husky && npm run depcheck --workspaces",
 		"check-licenses": "licensee --errors-only"
 	},
 	"repository": {
@@ -48,7 +48,6 @@
 		"@commitlint/config-conventional": "^20.0.0",
 		"@eslint/js": "^9.8.0",
 		"depcheck": "^1.4.7",
-		"docdash": "^2.0.2",
 		"eslint": "^9.35.0",
 		"eslint-config-google": "^0.14.0",
 		"eslint-plugin-ava": "^15.1.0",
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -24,9 +24,6 @@ jobs:
       with:
         node-version: 22.20.0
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -28,10 +28,8 @@
 		"lint:commit": "commitlint",
 		"unit": "npm run unit --workspaces",
 		"coverage": "npm run coverage --workspaces",
-		"jsdoc": "npm run jsdoc-generate && open-cli site/api/index.html",
-		"jsdoc-generate": "node ./scripts/isWorkspace.js && npm run jsdoc-generate-workspace || npm run jsdoc-generate-local",
-		"jsdoc-generate-local": "jsdoc -c ./jsdoc.json -t $(node -p 'path.dirname(require.resolve(\"docdash\"))') ./ || (echo 'Error during JSDoc generation! Check log.' && exit 1)",
-		"jsdoc-generate-workspace": "jsdoc -c ./jsdoc-workspace.json -t $(node -p 'path.dirname(require.resolve(\"docdash\"))') ../ || (echo 'Error during JSDoc generation! Check log.' && exit 1)",
+		"jsdoc": "npm run jsdoc-generate && open-cli packages/documentation/jsdoc/site/api/index.html",
+		"jsdoc-generate": "jsdoc -c packages/documentation/jsdoc/jsdoc.json -t $(node -p 'path.dirname(require.resolve(\"docdash\"))') ./ || (echo 'Error during JSDoc generation! Check log.' && exit 1)",
```
d3xter666: As all of the content related to JSDoc is now within the documentation package, maybe it makes sense those scripts also to be positioned there. That way, we could decouple from the root package. Eventually, those scripts can be referenced here if needed

maxreichmann: This was actually the original plan. But since we're using node_modules as sources, we would have to traverse the file hierarchy upwardly. This is not supported in jsdoc configs. I wrote about this in the PR description note.

RandomByte: Hm, doesn't the documentation package have it's own `node_modules`? At least for me that's the case.

Make sure to run npm install in the root **and** in the documentation package. Also move or add the required jsdoc dependencies to the documentation package.json

matz3: The `source/includePattern` seems to currently use the symlinked packages from the workspace, which is likely the reason the current setup only works if it is run from the root.

Using the local development state of the projects is however not intended. It would publish new APIs before they are available to consumers.

I think this has been discussed but not written down in the task. It aligns with the same requirement for the `ui5.yaml` schema, which should also only contain changes that have been released.

This can be achieved by declaring the npm dependencies in the `documentation` package. As it is not part of the workspace, it won't share the versions with the local workspace. With that, the script can also be moved into the `documentation` package to address the feedback of @d3xter666.

maxreichmann: Thanks to all for the tips.
With [c4e9cc5](https://github.com/UI5/cli/pull/1184/commits/c4e9cc5f30ad29f7b034cdc0f143afc21930744a), the scripts are now usable in `/packages/documentation`.

Also, the build output will now be at `/packages/documentation/dist/api/*` to be more aligned with the vitepress build.

PR #1187, opened 2025-11-07
refactor: REUSE tool setup in UI5 CLI repository

d3xter666:
JIRA: CPOUI5FOUNDATION-1150

- REUSE runs for every package, so that it's sure that each package compliance is correctly set and meets the criteria.
- REUSE for the whole monorepo, so a report can be generated. Root level files are checked at this point


.github/workflows/reuse-compliance.yml
```diff
@@ -12,10 +12,50 @@ on:
...
+  final-report:
     name: Compliance Check
-    runs-on: ubuntu-latest
+    needs: [packages-compliance-check]
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v5
-    - name: Execute REUSE Compliance Check
+    
+    - name: Temporarily add MIT license for full repo compliance
```
matz3: My understanding of the REUSE tool is that they are cloning the repo and run the checks on their infrastructure. Our CI integration is only to ensure we are not breaking the compliance, but AFAIK the `fsfe/reuse-action` does not report anything to their service. Only adding the license to the top-level folder will most probably not fulfil the REUSE compliance check that runs on their side.

d3xter666: I have removed the download. It was adding unnecessary overhead!

As we have discussed, now the reuse compliance runs for the individual packages as well as for the whole monorepo.

REUSE.toml
```diff
@@ -4,32 +4,23 @@ SPDX-PackageSupplier = "SAP OpenUI5 <openui5@sap.com>"
...
-SPDX-FileCopyrightText = ["2010 Sencha Inc.", "2011 LearnBoost", "2011 TJ Holowaychuk", "2014-2015 Douglas Christopher Wilson"]
-SPDX-License-Identifier = "MIT"
-
-[[annotations]]
-path = "packages/server/lib/middleware/testRunner/**"
-precedence = "aggregate"
-SPDX-FileCopyrightText = "2025 SAP SE or an SAP affiliate company and OpenUI5 contributors"
-SPDX-License-Identifier = "Apache-2.0"
-
+# Special exceptions - different copyright holder
```
matz3: The package-specific REUSE.toml files are aggregated when linting the whole repository. So I would assume that the root REUSE.toml file won't include duplicate exceptions for files within a package. The annotations for ui5-server have been removed, but for builder and documentation they are kept. Why is this necessary?

d3xter666: Done. Thanks

PR #1173, opened 2025-10-28
ci(release-please): Implement release please setup for UI5 CLI

d3xter666:
JIRA: CPOUI5FOUNDATION-1142

PR #1189, opened 2025-11-10
ci: Setup Coveralls for monorepo

maxreichmann:
JIRA: CPOUI5FOUNDATION-1153

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -37,7 +37,7 @@ jobs:
       run: npm run depcheck
 
     - name: Run unit tests
-      run: npm run coverage
+      run: npm run unit
 
     - name: Generate JSDoc documentation
       working-directory: packages/documentation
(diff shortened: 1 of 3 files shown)


.github/workflows/github-ci.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       run: npm run depcheck
 
     - name: Run unit tests
-      run: npm run coverage
+      run: npm run unit
```
d3xter666, line 40: You can run external gh action and wait to complete. 
Check this sample here: https://github.com/UI5/cli/pull/1173/files#diff-40a82b574505e06738cecbb56e70a40cb758eb7340d0531952d5d7e173f07769R38-R43

In this case, the coverage.yml can be triggered from here

maxreichmann: I considered this, but I'm not sure if we want to introduce so many dependencies within workflows. I personally prefer it more encapsulated for this regard (even though it's slower).

matz3: I think the separate workflow is fine, but with that we could question whether we actually need to run the tests in this workflow at all. As long as the coverage workflow also uses the minimum required Node.js version, I don't see a benefit of running the tests again in github-ci.yml.

PR #1194, opened 2025-11-17
docs(documentation): Add RFC 0019 Bundle-Info Preload

RandomByte:
**[Read the rendered markdown here](https://github.com/UI5/cli/blob/rfc-bundle-info-preload/rfcs/0019-bundle-info-preload.md)**

JIRA: CPOUI5FOUNDATION-1025

### DIFF
--- a/rfcs/0019-bundle-info-preload.md
+++ b/rfcs/0019-bundle-info-preload.md
@@ -64,6 +64,8 @@ To allow the UI5 runtime to identify libraries built with this new approach, the
 
 The presence of `"bundleVersion": 2` signals to the UI5 runtime that the library uses the "Bundle-Info Preload" format. The runtime can then safely and efficiently load the `library-preload.js` bundles for any lazy dependencies. If this flag is absent, the runtime can assume a legacy bundling format and issue a warning or error, as it expects all libraries for UI5 2.0+ to be built with a compatible UI5 CLI version.
 
+Note that this manifest attribute will only be added in a later manifest version. Therefore, the UI5 CLI must also adapt the manifest version accordingly when building libraries.
+
 ### Planned Runtime Behavior
 
 In a future UI5 framework release the following behavior might be implemented:
@@ -94,6 +96,8 @@ The "Bundle-Info Preload" feature is designed to work transparently for most dev
 ## Drawbacks
 
 UI5 releases prior to version 1.74 do not provide all required UI5 loader features, making the new bundling incompatible with these versions. Developers still targeting older releases (such as 1.71) will need to stick with an older release of UI5 CLI where the new bundling is not used.
+
+Similarly, the addition of the manifest flag requires the use of a specific manifest version which might not be supported by older UI5 releases. However, due to lack of validation of the manifest version in older UI5 releases, this is not expected to cause practical issues.
 
 Another drawback is the introduction of an additional request for the `_library-content.js` file. However, this request only occurs if the library is actually used, and it happens in parallel with other requests. For libraries that are declared but never used, this approach saves bandwidth by not loading the large content bundle at all. For eagerly used libraries, the performance impact is expected to be negligible and outweighed by the benefits of parallel loading and simplified development.
 
--- a/rfcs/0019-bundle-info-preload.md
+++ b/rfcs/0019-bundle-info-preload.md
@@ -2,12 +2,12 @@
 - RFC PR: [#1194](https://github.com/UI5/cli/pull/1194)
 - Issue: -
 - Affected components <!-- Check affected components by writing an "X" into the brackets -->
-    + [x] [ui5-builder](./packages/builder)
-    + [ ] [ui5-server](./packages/server)
-    + [ ] [ui5-cli](./packages/cli)
-    + [ ] [ui5-fs](./packages/fs)
-    + [ ] [ui5-project](./packages/project)
-    + [ ] [ui5-logger](./packages/logger)
+    + [x] [builder](./packages/builder)
+    + [ ] [server](./packages/server)
+    + [ ] [cli](./packages/cli)
+    + [ ] [fs](./packages/fs)
+    + [ ] [project](./packages/project)
+    + [ ] [logger](./packages/logger)
 
 
 # RFC 0019 Bundle Info Preload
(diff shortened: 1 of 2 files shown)


rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+
+1. `library-preload.js`: A small initial bundle containing the `library.js`, its direct dependencies, as well as "bundle-info" describing the content of the second bundle
+2. `_library-content.js`: A larger secondary bundle containing the remaining modules of the library.
+
+This allows the UI5 runtime to always load the small `library-preload.js` for all declared library dependencies, including those marked as "lazy". The larger `_library-content.js` is only loaded on-demand when a module from within it is actually requested. This new mechanism eliminates the need for application developers to manually trigger the loading of lazy dependencies.
+
+## Motivation
+
+Currently, UI5 applications often declare library dependencies as "lazy" in their `manifest.json` to improve initial loading times. However, this requires developers to explicitly load these libraries using `sap/ui/core/Lib#load("sap.m")` (or the deprecated variant `sap.ui.getCore().loadLibrary("sap.m")`) before any of their modules can be used. This is a manual and error-prone process.
```
matz3: It's `Lib.load` (static), and the name is passed via an options parameter. Maybe we should also link to the API Reference? https://ui5.sap.com/#/api/sap.ui.core.Lib%23methods/sap.ui.core.Lib.load.

I would rather prefer to not explicitly mention the deprecated variant, but I'm also ok with keeping it.

rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+By splitting the library preload, the essential part of a library can be loaded eagerly without a significant performance penalty. The UI5 loader can then automatically fetch the larger part of the library when it's first needed.
+
+This leads to the following benefits:
+- **Simplified Application Development:** Developers no longer need to manually manage the loading of lazy library dependencies.
+- **Improved Performance:** The large content bundle is only requested if the library is actually used, improving startup performance for applications with many optional dependencies.
+- **Efficient Parallel Loading:** This concept fits well with the "manifest-first" loading strategy, enabling more efficient and parallel loading of dependencies.
+
+## Detailed design
+
+The implementation of "Bundle-Info Preload" shall be handled by the UI5 CLI, specifically in the `ui5-builder` package.
```
matz3: The `ui5-*` naming convention doesn't appear anywhere anymore. Should we rather use `@ui5/builder` (npm package name) or just UI5 `builder` package instead?

rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+        "library": {
+            "bundleVersion": 2
+        }
+    }
+}
+```
+
+The presence of `"bundleVersion": 2` signals to the UI5 runtime that the library uses the "Bundle-Info Preload" format. The runtime can then safely and efficiently load the `library-preload.js` bundles for any lazy dependencies. If this flag is absent, the runtime can assume a legacy bundling format and issue a warning or error, as it expects all libraries for UI5 2.0+ to be built with a compatible UI5 CLI version.
+
+Note that this manifest attribute will only be added in a later manifest version. Therefore, the UI5 CLI must also adapt the manifest version accordingly when building libraries.
```
matz3, line 67: What manifest version would that be? A 2.x version? Can there be side-effects when a manifest version is updated during the build? What about a bump from 1.x to 2.x?

rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+
+### Planned Runtime Behavior
+
+In a future UI5 framework release the following behavior might be implemented:
+
+1. The application's `manifest.json` is loaded first.
+2. The UI5 runtime analyzes the library dependencies.
+3. For every library dependency (eager or lazy), it requests the corresponding `library-preload.js` file.
```
matz3: So where does the `bundleVersion` flag come into play at runtime?

rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+- **Content/Secondary Bundle:** The `_library-content.js` file.
+
+**Documentation:**
+- The UI5 CLI documentation for the `build` command should be updated to explain the new default bundling strategy.
+
+## Drawbacks
+
+UI5 releases prior to version 1.74 do not provide all required UI5 loader features, making the new bundling incompatible with these versions. Developers still targeting older releases (such as 1.71) will need to stick with an older release of UI5 CLI where the new bundling is not used.
+
+Similarly, the addition of the manifest flag requires the use of a specific manifest version which might not be supported by older UI5 releases. However, due to lack of validation of the manifest version in older UI5 releases, this is not expected to cause practical issues.
```
matz3: What about manifest version checks during deployment?

rfcs/0019-bundle-info-preload.md
```diff
@@ -0,0 +1,117 @@
...
+**Terminology:**
+- **Initial/Primary Bundle:** The new, smaller `library-preload.js`.
+- **Content/Secondary Bundle:** The `_library-content.js` file.
+
+**Documentation:**
+- The UI5 CLI documentation for the `build` command should be updated to explain the new default bundling strategy.
+
+## Drawbacks
+
+UI5 releases prior to version 1.74 do not provide all required UI5 loader features, making the new bundling incompatible with these versions. Developers still targeting older releases (such as 1.71) will need to stick with an older release of UI5 CLI where the new bundling is not used.
```
matz3: Does that mean that there needs to be a maintained UI5 CLI release compatible with UI5 1.71, as long as that UI5 version is supported?

rfcs/0019-bundle-info-preload.md
RandomByte: Thanks! Changed to `sap/ui/core/Lib.load({name: "sap.m"})`

rfcs/0019-bundle-info-preload.md
RandomByte: Changed to `builder` and updated the template as well 👍

rfcs/0019-bundle-info-preload.md
RandomByte: I can't really judge on that. Looping in @codeworrior, @loginger, @H4ze

rfcs/0019-bundle-info-preload.md
RandomByte: Added a bullet point for that 👍

rfcs/0019-bundle-info-preload.md
RandomByte: I'm not familiar with those so I didn't know. I updated the text to now basically say the opposite and point out the problem. Also added this to the unresolved questions

rfcs/0019-bundle-info-preload.md
RandomByte: Either that or we provide a switch to turn off the new bundling (which might also solve the manifest version problem). Updated the text.

rfcs/0019-bundle-info-preload.md
codeworrior: Yes, a 2.x version - the first 2.x version that lists the flag in the schema. For libraries, the schema version is not used as a switch at runtime, so I don't expect side effects when bumping it from 1.x to 2.x. 
BTW: for many libs (especially. all framework libs) the manifest is anyhow generated.

PR #1200, opened 2025-11-18
fix: Documentation build with NPM dependencies

d3xter666:
The isolated `documentation` package does not have its own `package-lock.json`, so the documentation publishing action fails on `npm ci` command.

### DIFF
--- a/internal/documentation/jsdoc/jsdoc.json
+++ b/internal/documentation/jsdoc/jsdoc.json
@@ -37,9 +37,9 @@
 	"openGraph": {
 		"title": "UI5 CLI - API Reference",
 		"type": "website",
-		"image": "https://ui5.github.io/cli/v5/images/UI5_logo_wide.png",
+		"image": "https://ui5.github.io/cli/v4/images/UI5_logo_wide.png",
 		"site_name": "UI5 CLI - API Reference",
-		"url": "https://ui5.github.io/cli/v5/"
+		"url": "https://ui5.github.io/cli/v4/"
 	},
 	"docdash": {
 		"sectionOrder": [


.github/workflows/deploy-vitepress-docs.yaml
```diff
@@ -49,7 +49,9 @@ jobs:
                   cp -r internal/documentation/. ${{ runner.temp }}/isolated-docs/
             - name: Install documentation dependencies from NPM
               working-directory: ${{ runner.temp }}/isolated-docs
-              run: npm ci --engine-strict
+              # npm ci won't work as this package relies on the monorepo's package-lock.json
+              # Run npm install instead as we want latest @ui5/* packages to be installed from NPM
+              run: npm install --engine-strict
```
matz3: This will fix the install, but we will also get the latest transitive versions for all dependencies.
What about declaring the packages under a different name in the documentation package.json  (e.g. `"@ui5/cli-latest": "npm:@ui5/cli@^4"`). Not sure if this works, but it could be a way of still only having one lockfile but that we can have different versions of the same package within our workspace.

RandomByte: Just a another idea: we might be able to use the shrinkwrap extractor here as well

d3xter666: With the shrinkwrap we need to move the `@ui5/*` dependencies from `devDependencies` to `dependencies` section and it won't hurt.

The npm aliases workaround seems to be working pretty well now.

d3xter666: I just verified that the npm alias solution builds well on [CI](https://github.com/d3xter666/cli/tree/gh-pages/v5/api) as well as verified locally that the correct version from NPM is being downloaded and no information from monorepo's packages is being used.

The thing with local verification is that NPM sometimes creates symlinks and sometimes creates a copy when there are cross dependencies within the monorepo.
In order to validate, it downloaded from NPM, I have changed some fields in the local `package.json` files and compared them among `@ui5/*` & `@ui5/*-npm` aliases

RandomByte: Sounds good, cool solution 👍

RandomByte:
LGTM, giving @matz3 a chance to review as well

internal/documentation/jsdoc/jsdoc-workspace.json
```diff
@@ -0,0 +1,71 @@
...
+				"include": ["jsdoc/templates/custom.css"]
+			}
+		}
+	},
+	"openGraph": {
+		"title": "UI5 CLI - API Reference",
+		"type": "website",
+		"image": "https://ui5.github.io/cli/v4/images/UI5_logo_wide.png",
+		"site_name": "UI5 CLI - API Reference",
+		"url": "https://ui5.github.io/cli/v4/"
```
matz3: The workspace state on "main" branch already now represents v5, isn't it?

d3xter666: Nice catch! Thanks!
I have also adjusted the `jsdoc.json`

matz3: Isn't `jsdoc.json` currently still used with the `v4` versions (the new npm dependencies in package.json)?

d3xter666: True! Thanks!
Reverted with https://github.com/UI5/cli/pull/1200/commits/bf904f29f7154eb22a8b8f4b37b8cb7dcbe253ab

PR #1205, opened 2025-11-20
ci: Fix Coveralls Bot from spamming coverage info

maxreichmann:
Before this PR, the Coveralls Bot's been spamming the Coverage info as PR comments. It's assumed, the bot's doing it because we have a Coveralls upload step for every package.

This PR tries to fix this by using the built-in "parallel" upload job feature by Coveralls itself. For now, this seems to have fixed the problem (as seen in the PR checks themselves).
This only creates one minor drawback: There are now 6 instead of only one check in the list.

---
Additionally, this PR changes the tests execution step (inside `github-ci.yml`) from "_npm run unit_" back to "_npm run coverage_".
This is done because `coverage.yml` only checks inside "packages/" and not "internal/".

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -47,6 +47,7 @@ jobs:
       with:
         parallel: true
         base-path: packages/${{ matrix.package }}
+        flag-name: ${{ matrix.package }}
 
   finish:
     needs: coverage


.github/workflows/coverage.yml
```diff
@@ -29,46 +38,22 @@ jobs:
...
       run: npm run coverage
 
-    - name: Send report to Coveralls for package @ui5/builder
+    - name: Send parallel reports to Coveralls for all packages
       uses: coverallsapp/github-action@v2.3.6
       continue-on-error: true # Do not fail the job if coverage reporting fails (e.g. service is down)
       with:
-        file: packages/builder/coverage/lcov.info
-        base-path: packages/builder
+        parallel: true
```
d3xter666, line 48: `flag-name` is required when this prop is endbled:

> flag-name | optional (unique required if parallel) | Job flag name, e.g. "Unit", "Functional", or "Integration". Will be shown in the Coveralls UI.

.github/workflows/coverage.yml
```diff
@@ -29,46 +38,22 @@ jobs:
...
-        base-path: packages/server
+  finish:
+    needs: coverage
+    if: ${{ always() }}
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Close parallel build
+        uses: coverallsapp/github-action@v2.3.6
+        with:
+          parallel-finished: true
```
d3xter666, line 60: I'm not sure about this, but shall we combine this with `carryforward`

maxreichmann: What I understood is that `carryforward` is only needed if you don't want to rebuild the entire project and reuse coverages from previous builds. But for us, we are "building" everything (every package) every time.

.github/workflows/coverage.yml
maxreichmann: Done in 6009c2c

maxreichmann:
*Rebased*

maxreichmann:
*Rebased*

PR #1206, opened 2025-11-21
ci: Enable schema build for monorepo

maxreichmann:
JIRA: CPOUI5FOUNDATION-1162

+ Schema build script moved to `internal/documentation/scripts/`
    - allows a parameter to set if schema should be based on local state or latest published npm state
+ npm scripts for this regard moved to `internal/documentation/`
+ Deployment CI jobs adjusted (but still disabled till final v5 release)
+ Changes in local workspace will be reflected in local build of schema
    - CI will generate schemas based on latest published npm version


internal/documentation/scripts/buildSchema-workspace.js
```diff
@@ -0,0 +1,54 @@
...
+import traverse from "traverse";
+
+try {
+	// Use ui5.yaml.json and ui5-workspace.yaml.json
+	const schemaNames = ["ui5", "ui5-workspace"];
+
+	schemaNames.forEach(async (schemaName) => {
+		// Using the local version of @ui5/project to use the current development version for local testing
+		const SOURCE_SCHEMA_PATH = fileURLToPath(
+			new URL(`./lib/validation/schema/${schemaName}.json`, import.meta.resolve("@ui5/project/package.json"))
```
matz3: Please [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). The two scripts only seem to differ in this line, so I'd propose to make the script configurable to work in both scenarios.

maxreichmann: Done in 5354155

PR #1211, opened 2025-11-24
refactor: Release all packages as grouped version

d3xter666:
Packages now always have the exact same version. The bump is based on the package with the highest value: https://github.com/googleapis/release-please/blob/009c2a61a038843957e7002aa1c32e317b654aa1/docs/manifest-releaser.md#linked-versions

JIRA: CPOUI5FOUNDATION-1166

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -34,6 +34,18 @@
 	"plugins": [
 		{
 			"type": "node-workspace"
+		},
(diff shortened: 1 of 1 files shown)


d3xter666:
Sample build: https://github.com/d3xter666/cli/pull/1

PR #1213, opened 2025-11-24
build: Update package-lock.json

d3xter666:
We have hit the following npm issue: https://github.com/npm/cli/issues/4828

Recreating the `package-lock.json` should fix that

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -71,6 +71,8 @@
 		},
 		"internal/documentation/node_modules/@types/node": {
 			"version": "22.19.1",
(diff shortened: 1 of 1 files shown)


d3xter666:
I have tested the solution on a fork on my account: https://github.com/d3xter666/cli/actions/runs/19635367626 and it seems to work well now.

I'd appreciate any better ideas

matz3 (approved):
Please rather use `build` as this is not fixing an issue on consumer side.

PR #1216, opened 2025-11-25
ci: Enable publishing to NPM

d3xter666:
JIRA: CPOUI5FOUNDATION-1166

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -50,9 +50,9 @@ jobs:
 
       - name: Publish ${{ matrix.package }} package
         working-directory: packages/${{ matrix.package }}
-        run: |
+        run: | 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -52,9 +52,7 @@ jobs:
         working-directory: packages/${{ matrix.package }}
         run: |
           echo "🚀 Publishing @ui5/${{ matrix.package }}"
-          
-          # TODO: Uncomment when ready to publish
-          # npm publish --access public
+          npm publish --access public
```
matz3: For the pre-releases, the npm dist tag should be `next` instead of the default `latest`, right?

d3xter666: Thanks!

PR #1165, opened 2025-10-21
feat(builder): Enhance generateFlexChangeBundle task to add new 'flexBundle' property to the manifest.json

d3xter666:
JIRA: CPOUI5FOUNDATION-1125

PR #1218, opened 2025-11-26
refactor(builder): generateFlexChangesBundle should not rely on existing manifest.json

d3xter666:
When running the `generateFlexChangesBundle` task, `manifest.json` might not be present.

Follow-up of https://github.com/UI5/cli/pull/1165


d3xter666:
Upport of https://github.com/SAP/ui5-builder/pull/1178

matz3 (approved):

RandomByte:
I thought upports should not appear in the changelog. Will we always modify the changelog or use a different prefix in the commit message?

d3xter666:
> I thought upports should not appear in the changelog. Will we always modify the changelog or use a different prefix in the commit message?

I think we discussed the `refactor` back then, didn't we?
However, with the current release either ways we'll wipe out everything, but the new component type and specVersion 5, so for this change it should be ok?

matz3:
https://github.com/UI5/cli/blob/v4/docs/Guidelines.md#implementing-fixes-across-v4-and-main-branches currently doesn't mention which prefix to use specifically. This should be enhanced.

PR #1220, opened 2025-11-27
fix: JSDoc build

d3xter666:
The JSDoc configuration was set to exclude specific files containing custom UI5 tags (like `@ui5-restricted`) that aren't recognized by JSDoc. However, **these exclusion patterns weren't working** in the monorepo setup.

With npm workspaces, dependencies can be installed in multiple locations:
- Root level: `../../node_modules/@ui5/` (relative to the documentation package)
- Package level: `./node_modules/@ui5/` (within the documentation package itself)

The original `jsdoc.json` only had exclusion patterns for the root-level path. When npm installed `@ui5/builder-npm` at the package level, the excluded files (like `JSTokenizer.js`) were still being processed. Since these files use the custom `@ui5-restricted` tag and `allowUnknownTags: false` was configured, JSDoc failed on encountering these unknown tags.

### DIFF
--- a/internal/documentation/jsdoc/jsdoc.json
+++ b/internal/documentation/jsdoc/jsdoc.json
@@ -3,12 +3,16 @@
 		"allowUnknownTags": false
 	},
 	"source": {
-		"include": ["jsdoc/index.md", "../../node_modules/@ui5"],
+		"include": ["jsdoc/index.md", "../../node_modules/@ui5", "./node_modules/@ui5"],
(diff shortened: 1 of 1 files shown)


d3xter666:
fixes: https://github.com/UI5/cli/actions/runs/19738946605/job/56557706752?pr=1198

d3xter666:
Closing in favour of: https://github.com/UI5/cli/pull/1221

PR #1223, opened 2025-11-28
docs(documentation): Migration guide v5

d3xter666:
JIRA: CPOUI5FOUNDATION-1166

Rendered document: https://github.com/UI5/cli/blob/a4e230a0b0a0880d852e26827ae8d4c68405a844/internal/documentation/docs/updates/migrate-v5.md

### DIFF
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -68,7 +68,7 @@ Update your `ui5.yaml` file to use the Component Type and Specification Version 
 
 Component Type follows a standardized directory structure:
 
-- **Move `webapp/test` to `test`** - Test files are now at the project root level
+- **Move `webapp/test` to `test`** - Test folder is now at the project root level
 - **Move `webapp` to `src`** - The source directory is now named `src` instead of `webapp`
 
 ::: code-group
@@ -107,6 +107,10 @@ my-app/
 
 The `index.html` file is typically moved from `/webapp` to `/test` since it's primarily used for testing the component.
 
+::: tip Note
+If your application needs an HTML page for purposes other than testing, this might be an indicator that the project should continue to use the **application type** instead of migrating to the component type.
+:::
+
 Update the bootstrap script path to the correct relative location (taking the project's namespace into account) and remove the obsolete `data-sap-ui-resource-roots` configuration:
 
 ```diff
@@ -123,6 +127,12 @@ Update the bootstrap script path to the correct relative location (taking the pr
       data-sap-ui-async="true">
   </script>
 ```
+
+::: tip Alternative: Using `<base>` Tag
+Instead of updating the bootstrap script path, you can add a `<base href="../../../../../">` tag in the `<head>` section of your HTML file. This allows you to keep the original path references.
+
+**Important:** The `<base>` tag affects **all relative URLs** in the document (including images, stylesheets, links, scripts, and even in-page anchors like `#some-id`). If certain resources are not loading correctly after adding the `<base>` tag, check whether they were using relative paths that are now being resolved differently than intended.
+:::
 
 #### 4. Adjust `test/testsuite.qunit.html`
 


internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,116 @@
...
+
+Try the alpha release in your projects via: `npm i --save-dev @ui5/cli@next`  
+Or update your global install via: `npm i --global @ui5/cli@next`
+
+**Note:** This is a pre-release version.
+:::
+
+## Breaking Changes
+
+- **All UI5 CLI Modules require Node.js >= 22.20.0 or >= 24.0.0**
```
matz3: 23 is not supported, but `>= 22.20.0` would allow it. The current range is `^22.20.0 || >=24.0.0`.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,116 @@
...
+
+- **@ui5/cli: `ui5 init` defaults to Specification Version 5.0**
+
+- **@ui5/project: New Component Type**
```
matz3: What is the breaking change w.r.t. the introduction of the new type? The type needs to be actively used (new feature). I don't see any incompatible changes when not actively changing something.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,116 @@
...
+
+	A new project type `component` has been introduced with **Specification Version 5.0**. Component Type projects follow standardized conventions:
+	- Source files in `/src` directory
+	- Test files in `/test` directory
+
+	See [RFC 0018](https://github.com/UI5/cli/blob/rfc-component-type/rfcs/0018-component-type.md) for details.
+
+## Node.js and npm Version Support
+
+This release requires **Node.js version v22.20.0 and higher or v24.0.0 and higher** as well as npm v8 or higher.
```
matz3: Explicitly mention that 23 is not supported (same as for v21 in the migrate-v4 guide)

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,116 @@
...
+  name: my.sample.app
+```
+:::
+
+#### 2. Restructure Project Directories
+
+Component Type follows a standardized directory structure:
+
+- **Move `/webapp` to `/src`** - The source directory is now named `src` instead of `webapp`
+- **Move `/webapp/test` to `/test`** - Test files are now at the project root level
```
matz3: When following those steps in order, the `webapp/test` folder will not exist anymore. I would suggest to first move out the test folder and then rename the webapp folder to src.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,116 @@
...
+│   ├── manifest.json
+│   ├── controller/
+│   └── view/
+└── test/
+    ├── index.html
+    ├── integration/
+    └── unit/
+```
+:::
+
```
matz3, line 105: Are those really all steps for a migration?

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Done

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Done

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Removed it

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Done

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Now should be better

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -11,21 +11,14 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 ## Breaking Changes
 
-- **All UI5 CLI Modules require Node.js >= 22.20.0 or >= 24.0.0**
+- **All UI5 CLI Modules require Node.js >= ^22.20.0 || >=24.0.0**
```
matz3:
```suggestion
- **All UI5 CLI Modules require Node.js ^22.20.0 || >=24.0.0**
```

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -109,6 +102,87 @@ my-app/
...
+:::
+
+**Changes:**
+- Remove the `data-sap-ui-testsuite` attribute
+- Remove the `data-sap-ui-resourceroots` attribute
+- Update the `src` path to the correct relative location
+
+#### 5. Delete Obsolete Test Files
+
+Delete the custom `Test.qunit.html` file from your test directory. This file is no longer needed. The framework-provided test page can now be used directly.
```
matz3: Only removing the file is not sufficient. The testsuite configuration also needs to be adjusted. See the changes of the sample-app: https://github.com/UI5/sample-app/compare/main...component-type

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -109,6 +102,87 @@ my-app/
...
+Update the bootstrap script path and remove the obsolete `data-sap-ui-resourceroots` configuration:
+
+::: code-group
+```html [Before (webapp/index.html)]
+<script
+    id="sap-ui-bootstrap"
+    src="resources/sap-ui-core.js"
+    data-sap-ui-libs="sap.m"
+    data-sap-ui-theme="sap_horizon"
+    data-sap-ui-resourceroots='{
```
matz3: outdated spelling

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I have updated with the things in there. Please, let me know if I've missed something else

d3xter666 (commented):
LGTM

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,195 @@
...
++ type: component
+  metadata:
+    name: my.sample.app
+```
+
+#### 2. Restructure Project Directories
+
+Component Type follows a standardized directory structure:
+
+- **Move `webapp/test` to `test`** - Test files are now at the project root level
```
codeworrior: [minor] I would rather say "...test folder...", not "...test files...". The files aren to at project root level.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,195 @@
...
+└── test/
+    ├── index.html
+    ├── integration/
+    └── unit/
+```
+:::
+
+#### 3. Adjust `test/index.html`
+
+The `index.html` file is typically moved from `/webapp` to `/test` since it's primarily used for testing the component.
```
codeworrior, line 108: If an app needs an HTML page for purposes other than testing, this might be an indicator that the project rather should continue to use the application type, isn't it?

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -0,0 +1,195 @@
...
+
+Update the bootstrap script path to the correct relative location (taking the project's namespace into account) and remove the obsolete `data-sap-ui-resource-roots` configuration:
+
+```diff
+  <script
+      id="sap-ui-bootstrap"
+-     src="resources/sap-ui-core.js"
++     src="../../../../../resources/sap-ui-core.js"
```
codeworrior, line 120: I wonder whether we shouldn't recommend the use of a `<base href="...">` tag now.

d3xter666: Yes, we can provide a hint that `<base...\>` can be used, but if we add this as recommendation, it could lead to mess with the paths in case of any other relative resources are being used

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Done

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Haven't thought about this! Thanks!
I'll add this directly as a note

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I have added this as informtaion to the document

codeworrior (commented):
Looks good to me.

But I would like to hear @matz3's opinion on the index.html note (I remember him to name the need for an HTML page as one possible differentiator between application and component).

Plus, after reading the lengthy notes about the `<base>` tag, I wonder whether me proposing it really improves the guide or rather makes it too complex. So I don't mind if you decide to keep it rather simple. Experienced developers might anyhow decide to use a base tag.

internal/documentation/docs/updates/migrate-v5.md
matz3: I agree. For now, we should not recommend to switch to component type in case the application is started standalone in productive scenarios.

One idea @RandomByte and I had was to keep the `index.html` in `src` and get the same output with `--output-style Flat`. But this would require a (server-)absolute path in the HTML, as it would otherwise not point to the correct place. I think this is nothing we should recommend right now, but at least keep in mind.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -107,6 +107,10 @@ my-app/
 
 The `index.html` file is typically moved from `/webapp` to `/test` since it's primarily used for testing the component.
 
+::: tip Note
+If your application needs an HTML page for purposes other than testing, this might be an indicator that the project should continue to use the **application type** instead of migrating to the component type.
```
matz3, line 111: The example we used here is based on the UI5/sample-app, which rather falls into the category of of being a standalone UI5 app. Should we rather update the guidelines based on a typical Fiori application project?

RandomByte: Maybe we could make it clear that `index.html` only needs to be moved if it's actually present.

internal/documentation/docs/updates/migrate-v5.md
RandomByte: Are there any drawbacks to using `<base>`? Without knowing it too well, it seems like the better solution to me

d3xter666: In terms of migration, to me it seems the cleanest solution. As all of the resources were previously loaded from the destination now defined in the <base/> tag. 

However, adding new resources or editing resources might be a bit confusing, since the file is located at one destination and the tag points to a different one. And this is usually not obvious at first glance.

PR #1225, opened 2025-12-02
build: Apply npm package fixes

matz3:
Applies changes from "npm pkg fix".

### DIFF
--- a/internal/shrinkwrap-extractor/package.json
+++ b/internal/shrinkwrap-extractor/package.json
@@ -11,11 +11,12 @@
 	"license": "Apache-2.0",
 	"main": "lib/convertPackageLockToShrinkwrap.js",
 	"bin": {
-		"shrinkwrap-extractor": "./cli.js"
+		"shrinkwrap-extractor": "cli.js"
 	},
 	"repository": {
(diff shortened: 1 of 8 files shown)


internal/shrinkwrap-extractor/package.json
```diff
@@ -11,11 +11,11 @@
 	"license": "Apache-2.0",
 	"main": "lib/convertPackageLockToShrinkwrap.js",
 	"bin": {
-		"shrinkwrap-extractor": "./cli.js"
+		"shrinkwrap-extractor": "cli.js"
 	},
 	"repository": {
 		"type": "git",
-		"url": "git@github.com:UI5/cli.git"
+		"url": "git+ssh://git@github.com/UI5/cli.git"
```
flovogt: directory is missing

matz3: I would rather propose to remove most of the content from the internal packages as it is not relevant for them. We will never publish them.

But for now I can add the directory here, although e.g. the documentation package doesn't contain a `repository` field at all.

RandomByte (approved):

PR #1228, opened 2025-12-04
fix: Improve shrinkwrap generation handling

d3xter666:
Enhances: https://github.com/UI5/cli/pull/1231


internal/shrinkwrap-extractor/lib/convertPackageLockToShrinkwrap.js
```diff
@@ -100,6 +103,28 @@ export default async function convertPackageLockToShrinkwrap(workspaceRootDir, t
...
+ * 	- packages/cli/node_modules/foo -> node_modules/foo
+ * 	- packages/fs/node_modules/bar -> node_modules/bar
+ *
+ * @param {string} location - Package location from arborist
+ * @returns {string} - Normalized location for npm-shrinkwrap.json
+ */
+function normalizePackageLocation(location) {
+	// Remove workspace prefix (packages/*/...) and convert to standard node_modules path
+	// Match: packages/<workspace-name>/node_modules/... or packages/<workspace-name>
+	const workspacePattern = /^packages\/[^/]+\/(node_modules\/.+)$/;
```
matz3: Only removing the prefix won't solve the placement of a package in multiple versions, right?
I'm pushing a change to place the packages in a nested `node_nodules` directory.

I haven't tested it so far, but installing the same package in different major versions in `@ui5/cli` and `@ui5/fs` showed an expected result. Before my change, only one version was included in the lockfile.

matz3: I think there's still something wrong. In the fixtures, some modules are not contained in a version they are required by other packages, e.g. `ansi-regex": "^6.0.1"`. It's only contained in version 5.

RandomByte:
To verify the shrinkwrap:

```sh
npm i
cd packages/cli
node ../../internal/shrinkwrap-extractor/cli.js ../../
rm -rf node_modules # Delete node_modules within CLI package

# <Manually remove devDependencies in package.json>

npm ci --workspaces false
```

Before Matthias' fix this yielded the following error:

```
npm error Missing: @npmcli/package-json@7.0.4 from lock file
npm error Missing: @npmcli/package-json@7.0.4 from lock file
npm error Missing: globby@15.0.0 from lock file
npm error Missing: proc-log@6.1.0 from lock file
npm error Missing: proc-log@6.1.0 from lock file
npm error Missing: proc-log@5.0.0 from lock file
npm error Missing: proc-log@5.0.0 from lock file
npm error Missing: unicorn-magic@0.3.0 from lock file
```

Those dependencies where missing from the shrinkwrap, likely because multiple workspace packages dependent on different versions, overwriting each other.

I'm currently verifying the state after Matthias' fix. I suspect there might still be a problem with dependencies hoisted to the workspace root for packages other than the CLI, conflicting with dependencies of the CLI.

RandomByte:
> I suspect there might still be a problem with dependencies hoisted to the workspace root for packages other than the CLI, conflicting with dependencies of the CLI.

I added a failing test for this case. I'll work on a fix

internal/shrinkwrap-extractor/lib/convertPackageLockToShrinkwrap.js
d3xter666: Done

internal/shrinkwrap-extractor/test/expected/package.c/npm-shrinkwrap.json
```diff
@@ -0,0 +1,52 @@
...
+				"@ui5/module-b": "^1.0.0"
+			},
+			"devDependencies": {}
+		},
+		"node_modules/@sapui5/some-thirdparty": {
+			"version": "1.0.0",
+			"resolved": "https://registry.npmjs.org/package/version.tgz",
+			"integrity": "sha512-mock-integrity-hash"
+		},
+		"node_modules/@ui5/target/node_modules/@sapui5/some-thirdparty": {
```
RandomByte: I don't think this will work. If `@ui5/target` is the root package, it will never search the path `node_modules/@ui5/target/node_modules/` for its dependencies.

For module `@sapui5/some-thirdparty`, the target package would resolve it to `node_modules/@sapui5/some-thirdparty`, which contains the wrong version now...

RandomByte (changes requested):

d3xter666:
I just rebased from main and pushed to resolve conflicts and this PR simply closed

PR #1230, opened 2025-12-08
ci: Add internal benchmark tool

matz3:
See: https://github.com/UI5/cli/blob/main/internal/benchmark/README.md

JIRA: CPOUI5FOUNDATION-1181

### DIFF
--- a/internal/benchmark/lib/services/HyperfineRunner.js
+++ b/internal/benchmark/lib/services/HyperfineRunner.js
@@ -89,10 +89,10 @@ export default class HyperfineRunner {
 				stdio: "inherit",
 				captureOutput: false,
 				env: {
+					// Inherit all other environment variables
+					...process.env,
 					// Disable invocation of local UI5 CLI installations
 					UI5_CLI_NO_LOCAL: "X",
-					// Inherit all other environment variables
-					...process.env
 				},
 				errorMessage: "hyperfine exited with non-zero code"
 			});


internal/benchmark/package.json
```diff
@@ -0,0 +1,24 @@
...
+	"license": "Apache-2.0",
+	"type": "module",
+	"bin": {
+		"ui5-cli-benchmark": "./cli.js"
+	},
+	"scripts": {
+		"test": "npm run lint && npm run coverage && npm run depcheck",
+		"unit": "node --test 'test/e2e/**/*.js'",
+		"unit-watch": "node --test --watch 'test/e2e/**/*.js'",
+		"coverage": "node --test --experimental-test-coverage 'test/e2e/**/*.js'",
```
d3xter666, line 13: As we use `lcov` to report test coverage to Coveralls, maybe it's worth to enable it also here by adding:
`--test-reporter=spec --test-reporter-destination=stdout --test-reporter=lcov --test-reporter-destination=coverage/lcov.info`

This will enable lcov as well as stdout for coverage.

internal/benchmark/lib/services/HyperfineRunner.js
```diff
@@ -0,0 +1,147 @@
...
+			// Execute hyperfine
+			await this.#spawnProcess("hyperfine", args, {
+				cwd: workingDirectory,
+				stdio: "inherit",
+				captureOutput: false,
+				env: {
+					// Disable invocation of local UI5 CLI installations
+					UI5_CLI_NO_LOCAL: "X",
+					// Inherit all other environment variables
+					...process.env
```
d3xter666: If we want to always avoid local CLI installation, then those should be inversed.
It's might be that `process.env` has `UI5_CLI_NO_LOCAL = null` and to overwrite it.

internal/benchmark/lib/benchmark/BenchmarkSpec.js
```diff
@@ -0,0 +1,122 @@
...
+			throw new Error("Benchmark configuration must be an object");
+		}
+		if (!config.command || typeof config.command !== "string") {
+			throw new Error("Benchmark must have a command string");
+		}
+		if (config.prepare !== undefined && typeof config.prepare !== "string") {
+			throw new Error("Benchmark prepare must be a string if provided");
+		}
+		if (!config.groups || typeof config.groups !== "object" || Object.keys(config.groups).length === 0) {
+			throw new Error("Benchmark must belong to at least one group");
```
d3xter666, line 26: This message is a bit confusing. I would interpret this as we do a validation agains the groups.
It's rather that no groups are deinfed

matz3: It is the validation of one benchmark specification. It must have at least one assigned group, so that the results are shown.

The validation of groups is done beforehand, within `Configuration.js`.
And a validation of whether the assigned groups actually exist, is also done there (right after the `BenchmarkSpec` instance is created).

internal/benchmark/lib/services/HyperfineRunner.js
matz3: That's right, thanks!

internal/benchmark/package.json
matz3: I agree, but are those results currently uploaded to coveralls (e.g. for the shrinkwrap-extractor)?

internal/benchmark/lib/benchmark/BenchmarkSpec.js
d3xter666: Yep, I saw that validation there, but here, standalone, to me it seems confusing

internal/benchmark/package.json
d3xter666: Nope, they are not, but I was wondering whether we want to be consistent.
Currently, we have couple of different test frameworks- `ava` and `nodejs` native and at least results could have been consolidated if we decide to include those in the overal reporting

matz3: Yes I agree and I don't think there's a reason against also reporting the coverage data of internal packages. I just wouldn't do it within this PR.

PR #1244, opened 2025-12-11
ci: Do not fail ci if Coverage reporting fails

maxreichmann:
As seen in this [GH Actions run](https://github.com/UI5/cli/actions/runs/20137761530/job/57796474383?pr=1243), the Coveralls servers can be down very easily. 

This adds `continue-on-error` to the coverage upload step too (previously only on coverage calculation steps)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -56,5 +56,6 @@ jobs:
     steps:
       - name: Close parallel build
         uses: coverallsapp/github-action@v2.3.7
+        continue-on-error: true # Do not fail the job if coverage reporting fails (e.g. service is down)
         with:
           parallel-finished: true


.github/workflows/coverage.yml
```diff
@@ -56,5 +56,6 @@ jobs:
     steps:
       - name: Close parallel build
         uses: coverallsapp/github-action@v2.3.7
+        continue-on-error: true # Do not fail the job if coverage reporting fails (e.g. service is down)
```
d3xter666, line 59: Do we really wanna do this?
We might end up in a situation where the coverage decreased, but it was not reported and no one will notice this!
In such case, some other, next run might fail and we will need to oinvestigate something that didn't come form the next failing (due to lower coverage) change

maxreichmann: For the coverage, we currently run a step for calculating each package's coverage. This already has the ` continue-on-error` flag.
Now, this PR only adds the same flag to the step after it (collecting all coverages and upload them bundled together).

So the risk you're describing is already in our CI. We can open-up the discussion again on moving away from the official Coveralls servers and hosting it ourselves.

d3xter666: I had the impression that this step posts the information to the Coveralls service, then Coveralls aggregates the information and posts it as a comment in the PR

maxreichmann: Yes you're right. This is the step after each coverage calculation. The report gets bundled together, uploaded and used to post the bot comment.

But, if we already have the flag in the step beforehand. Why not in this step?

d3xter666: If the Coverals service is down and this flag is set, then we can easily miss that the Coverage chnaged in a negative aspect- the job will be green. 
If we merge this potential change (with the bad coverage), all of the next changes where Coveralls work will start failing due to insufficient coverage. But this will not be introduced from the changes themselves. It will be comming from the main branch.

This is the hypotetical situation I see with this flag now

maxreichmann:
**Outcome of discussion** (Dec 15, @d3xter666 & @maxreichmann):

Since we don't want to miss if the Coverage actually went down and get notified about it, the approach is as follows:
We're keeping the flag for the parallel steps ("_Coverage / Check and Upload (packages/...)_") and not adding it to the upload step ("_Coverage / finish_"). When the servers are up, we can see the coverage percentage without problems. When the servers are down, we can easily calculate the coverage locally and are still be able to merge.

Closing.

PR #1249, opened 2025-12-15
[INTERNAL] Docs: Prevent updating 404.html

matz3:
An updated version of the 404.html will be deployed from the `main` branch
via https://github.com/UI5/cli/pull/1250.


scripts/buildDocs.sh
```diff
@@ -17,7 +17,7 @@ if [[ $MIKE_VERSION ]]; then # Build with Mike (versioning)
 	echo "Starting building & versioning Docs with Mike version: ${MIKE_VERSION}; alias: ${MIKE_ALIAS}..."
 	docker run --rm -v $(pwd):/docs --entrypoint mike \
 		--env GIT_COMMITTER_NAME="${GIT_COMMITTER_NAME}" --env GIT_COMMITTER_EMAIL="${GIT_COMMITTER_EMAIL}"  \
-		$DOCKER_IMAGE deploy $MIKE_VERSION $MIKE_ALIAS --rebase --update-aliases
```
d3xter666, line 20: `--update-aliases` enforces alias update. However, we don't have `next` defined [here](https://github.com/UI5/cli/pull/1249/files#diff-87e860af4b9db6c503ed680b6edb6333c6f8d7659f7d1a779a2487466bbc50f6L16) as alias and it still updates it.

This makes me think that the core reason behind this is [gh-pages/versions.json](https://github.com/UI5/cli/blob/gh-pages/versions.json#L7) where is the only place `next` alias is related to `v4`. Although Mike Docs state this file is for version selection only. To be on tghe safe side, please also remove it from there. It would require additional PR for `gh-pages`

Current solution should work as it will not enforce alias updates.
But when we enable v5 as `stable`, this one might break as there would be collisons.
I can't prove that, because I cannot think of a safe way to test it, but this is my interpretation

matz3: Good point. I also wasn't sure about this. So what about manually adjusting the `versions.json` in addition to this change?

d3xter666: Let's try the following to verify it. Let's update the `versions.json` in `gh-pages`. It's either ways incorrect!
Then we can see whether we need this change at all by retriggering the latest `deploy docs` action

matz3: I've updated this PR to only remove the 404.html, as it will now be handled via https://github.com/UI5/cli/pull/1250

PR #1259, opened 2025-12-17
[INTERNAL] Add message for v3 deprecation

d3xter666:
JIRA: CPOUI5FOUNDATION-875

### DIFF
--- a/overrides/main.html
+++ b/overrides/main.html
@@ -1,0 +1,12 @@
+{#-
+  This file was initially copied from https://squidfunk.github.io/mkdocs-material/customization/#overriding-blocks
+  It displays a message for the non-default branch
+-#}
+{% extends "base.html" %}
+
+{% block announce %}
+  ⚠️ UI5 CLI v3 has been deprecated! We highly recommend migrating to
+  <a href="{{ base_url ~ '/../stable/' }}">
+  <strong>the latest version.</strong>
+  </a>
+{% endblock %}


README.md
```diff
@@ -10,9 +10,9 @@
 
 > [UI5 CLI v4](https://sap.github.io/ui5-tooling/v4) is the latest and stable version 🎉
 >
-> [UI5 CLI v3](https://sap.github.io/ui5-tooling/v3) is a stable version and in maintenance mode 🚢
+> [UI5 CLI v3](https://sap.github.io/ui5-tooling/v3) has been deprecated 🚫
 >
-> [UI5 CLI v2](https://sap.github.io/ui5-tooling/v2) has been deprecated 🚫  
+> [UI5 CLI v2](https://sap.github.io/ui5-tooling/v2) has been deprecated 🚫
 >
 > Please migrate your projects to [UI5 CLI v3](https://sap.github.io/ui5-tooling/v3/updates/migrate-v3/)!
```
matz3: This isn't valid anymore, right?

I think we should try to reduce the information in the README, especially for older branches. They will become outdated and it's not realistic to maintain all of them for every new major version. I would propose to just not state anything here and refer to https://ui5.github.io/cli/stable/

PS: There are now two PRs for the `v3` branch with different commit messages. Wouldn't it be easier to combine them, as they are about the same purpose?

d3xter666: Done

PR #1262, opened 2025-12-18
refactor(builder): Use only /resources of sap.ui.core for library preload generation

RandomByte:
Adjust the glob pattern to exclude test-resources


RandomByte:
The wide pattern currently leads to excessive invalidation of the task during builds with the Incremental Build PoC. For example a change in a test-resource invalidates the task even though the resources are seemingly never used in sap.ui.core.

@codeworrior do you recall the reason for this pattern?

PR #1238, opened 2025-12-10
feat: Incremental Build

RandomByte:
Implementation of [RFC 0017 Incremental Build](https://github.com/UI5/cli/pull/1036)

This PR supersedes previous PoC: https://github.com/UI5/cli/pull/1199  

JIRA: CPOUI5FOUNDATION-1174

PR #1267, opened 2026-01-07
feat: Incremental Build

RandomByte:
Implementation of [RFC 0017 Incremental Build](https://github.com/UI5/cli/pull/1036)

This PR supersedes previous PoC: https://github.com/UI5/cli/pull/1238

JIRA: CPOUI5FOUNDATION-1174


packages/project/test/fixtures/application.a/custom-tasks/custom-task-2.js
```diff
@@ -0,0 +1,29 @@
...
+	console.log("Custom task 2 executed");
+
+	// For #1 Build: Read a file which is an input of custom-task-1 (which sets a tag on it)
+	const testJS = await workspace.byPath(`/resources/${projectNamespace}/test.js`);
+
+	// For #3 Build: Read a different file (which is NOT an input of custom-task-1)
+	const test2JS = await workspace.byPath(`/resources/${projectNamespace}/test2.js`);
+
+	const tag = taskUtil.getTag(testJS, taskUtil.STANDARD_TAGS.IsDebugVariant);
+	if (!test2JS && testJS) {
```
RandomByte: Why is this if necessary? Both bodies seem identical

packages/project/test/fixtures/application.a/custom-tasks/custom-task-0.js
```diff
@@ -0,0 +1,29 @@
...
+		// For #1 Build:
+		if (tag) {
+			throw new Error("Tag set by custom-task-1 is present in custom-task-0, which is UNEXPECTED.");
+		} else {
+			console.log("Tag set by custom-task-1 is not present in custom-task-0, as EXPECTED.");
+		}
+	} else {
+		// For #3 Build (NEW behavior expected as in #1):
+		if (tag) {
+			console.log("Tag set by custom-task-1 is present in custom-task-0 now, as EXPECTED.");
```
RandomByte: This is not expected. A task must never see tags only set by later tasks

packages/project/test/fixtures/application.a/custom-tasks/custom-task-2.js
maxreichmann: Yeah, my only motive behind this was to have two seperate logic blocks (one for each build) and to have it consistent with custom-task-0.js (for easier understanding). But I'll refactor it to only one.

packages/project/test/fixtures/application.a/custom-tasks/custom-task-0.js
maxreichmann: The tag for test.js is set by custom-task-1 in #1 Build.
In, #3 Build we then only touch test2.js, which skips execution of custom-task-1 (it's not an input of it). But  should custom-task-0 still not be able to see the tag? What's then the purpose of the cache?

RandomByte: We need to make sure that tasks always behave the same, no matter whether _other_ tasks are skipped (by using a valid cache) or not.

This means that a custom-task-0, which can't find a tag for resource `test.js` in the first build, _**must never**_ find a tag for that resource in any other build.

Otherwise you could end up with different build results, depending on whether you build your project with or without cache. This must to be avoided.

maxreichmann: Got it now. That's definitely the right way.
I'll refactor the tests now.

packages/project/test/fixtures/application.a/custom-tasks/custom-task-2.js
maxreichmann: Done in 03d0f0e96ce3d87b6dbe6a033a01827fa03af7a5

packages/project/test/fixtures/application.a/custom-tasks/custom-task-0.js
maxreichmann: Done in 03d0f0e96ce3d87b6dbe6a033a01827fa03af7a5

packages/project/test/fixtures/application.a/task.dependency-change.js
```diff
@@ -0,0 +1,36 @@
...
+				return;
+			}
+			if (projectsVisited.has(projectName)) {
+                return;
+            }
+            projectsVisited.add(projectName);
+            const project = taskUtil.getProject(projectName);
+			const newLibraryFile = await project.getReader().byGlob("**/newLibraryFile.js");
+			if (newLibraryFile.length > 0) {
+				console.log('New Library file found. We are in #4 build.');
```
RandomByte: Let's use the UI5 Logger via the `log` parameter above. E.g. `log.verbose("foo")`

packages/project/test/fixtures/application.a/custom-tasks-2/custom-task-1.js
```diff
@@ -0,0 +1,25 @@
...
+		console.log("Flag NOT set -> We are in #1 Build still");
+		buildRanOnce = true;
+		const tag = taskUtil.getTag(testJS, taskUtil.STANDARD_TAGS.IsDebugVariant);
+		if (!tag) {
+			throw new Error("Tag set during #1 Build is not readable, which is UNEXPECTED.");
+		}
+	} else {
+		console.log("Flag set -> We are in #2 Build");
+		const tag = taskUtil.getTag(testJS, taskUtil.STANDARD_TAGS.OmitFromBuildResult);
+		if (!tag) {
```
RandomByte, line 28: We should additionally validate that the `IsDebugVariant` tag **is not set** for the given resource in this case (second run).

maxreichmann:
*Rebased onto origin/main*

internal/e2e-tests/package.json
```diff
@@ -0,0 +1,23 @@
...
+		"unit": "node --test 'test/**/*.js'",
+		"unit-watch": "node --test --watch 'test/**/*.js'",
+		"coverage": "node --test --experimental-test-coverage 'test/**/*.js'",
+		"lint": "eslint ."
+	},
+	"devDependencies": {
+		"eslint": "^9.39.1"
+	},
+	"dependencies": {
+		"adm-zip": "^0.5.17"
```
maxreichmann: uses MIT license -> TODO: refactor tests to use different library / approach

RandomByte:
Have you considered adding the e2e tests in a dedicated PR for main? This way we could verify that they work correctly without the incremental build. It would also reduce the size of this PR.

internal/e2e-tests/package.json
RandomByte: MIT is usually fine, what is the problem here?

RandomByte:
Will be merged via https://github.com/UI5/cli/pull/1399

There are several known bugs that will be addressed in follow-ups before a new pre-release should be created.

PR #1278, opened 2026-01-19
docs: Escape HTML tags and replace pre in jsdoc examples

werwolf2303:


packages/builder/lib/processors/versionInfoGenerator.js
```diff
@@ -366,14 +365,15 @@ const processLibraryInfo = async (libraryInfo) => {
  * @param {string} parameters.options.rootProjectVersion Version of the root project
  * @param {module:@ui5/builder/processors/versionInfoGenerator~LibraryInfo[]} parameters.options.libraryInfos Array of
  *				objects representing libraries,
- *  e.g. <pre>
+ *  e.g.
+ *  @example
```
codeworrior: Is this correct? To my understanding, using a block tag like `@example` decouples the example from the parameter description. The lonely "e.g." then might confuse readers.

werwolf2303: True. It was indeed an issue. I replaced it with a Markdown style code block

packages/builder/lib/processors/jsdoc/lib/ui5/plugin.js
```diff
@@ -436,11 +436,12 @@ function getPropertyKey(prop) {
  *
  * If a <code>defaultKey</code> is given and if the <code>node</code> is not an object literal
  * but another simple type literal, the value is treated as a shortcut for
- * <pre>
+ * ```
```
matz3: This file is actually a copy of https://github.com/UI5/openui5/blob/master/lib/jsdoc/ui5/plugin.js and I would not expect that we include it into our JSDoc output.
We should not modify the file, and if not done already, we should exclude `packages/builder/lib/processors/jsdoc/lib/**` from the JSDoc build.

matz3 (approved):

PR #1287, opened 2026-01-26
feat: PoC: node modules & webcomponents integration

d3xter666:
JIRA: CPOUI5FOUNDATION-1196
JIRA: CPOUI5FOUNDATION-979

Implementations, researches and PoCs are based on the input from  [ui5-tooling-modules](https://www.npmjs.com/package/ui5-tooling-modules)

RFC: https://github.com/UI5/cli/blob/d7d7c1e4460a28815597a1df543eda0cfa83464a/rfcs/0020-npm-package-integration.md

Sample app build & serve: `internal/npm-integration-poc/consolidated-app/`

Simple rollup tests and builds: `internal/npm-integration-poc/scenarios/`

Shared code (`utils`): `internal/npm-integration-poc/lib/`


rfcs/0014-npm-package-integration.md
```diff
@@ -0,0 +1,654 @@
...
+- Issue: -
+- Affected components
+    + [ ] [ui5-builder](../packages/builder)
+    + [ ] [ui5-server](../packages/server)
+    + [ ] [ui5-cli](../packages/cli)
+    + [ ] [ui5-fs](../packages/fs)
+    + [ ] [ui5-project](../packages/project)
+    + [ ] [ui5-logger](../packages/logger)
+
+# RFC 0014 NPM Package Integration for UI5 Projects
```
RandomByte: Just a side note I just noticed: To avoid confusion, this RFC should use the number `0020`.

0014 is already taken for https://github.com/UI5/cli/pull/897  
0019 for https://github.com/UI5/cli/pull/1194

d3xter666: Done

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+5. **Modern Development** — Bridging UI5's AMD system and the modern JavaScript ecosystem
+
+### 2.3 Expected Outcome
+
+- **Out-of-the-Box Integration** — NPM packages work without configuration
+- **Build-Time Transformation** — Automatic bundling during `ui5 build`
+- **Dev-Time Support** — On-demand bundling during `ui5 serve` (deferred to [RFC 0017 Incremental Build](https://github.com/UI5/cli/blob/rfc-incremental-build/rfcs/0017-incremental-build.md))
+- **Standard Compliance** — Output follows UI5's AMD module format and naming conventions
+- **Web Component Support** — First-class `@ui5/webcomponents` integration with UI5 control wrappers
+- **Performance Optimization** — Package independence enables deduplication and browser caching
```
RandomByte: Not sure what this optimization refers to. Does it mean optimizing what ui5-tooling-modules does? Maybe this can be removed from this chapter?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+- **UI5 Application Source** — The starting point is the developer's application code. Controllers reference NPM packages via AMD dependencies (e.g., `sap.ui.define(['thirdparty/chart.js', ...])`), and XML views reference them via xmlns declarations (e.g., `xmlns:webc="thirdparty.@ui5.webcomponents"`). These `thirdparty/*` references are the convention that triggers the NPM integration pipeline.
+
+- **Phase 1: Integrated Dependency Analysis** — The existing `JSModuleAnalyzer` and `XMLTemplateAnalyzer` are extended to detect `thirdparty/*` patterns during their regular AST traversal. `espree.parse` walks the AST looking for `thirdparty/*` in AMD dependency arrays and ESM import statements, while `XMLTemplateAnalyzer` scans for `xmlns:thirdparty.*` declarations. `ResourceCollector` aggregates all detected NPM package names across all analyzed resources. **Key architectural decision**: This piggybacks on the AST parsing that already occurs for UI5 dependency analysis — no separate scanning phase is needed, eliminating duplicate parsing overhead. Output: a deduplicated set of NPM package names (e.g., `['chart.js', 'react', 'react-dom', '@ui5/webcomponents']`).
```
RandomByte: This `thirdparty/*` naming convention is only introduced in this concept, right? `ui5-tooling-modules` seems to simply use the npm package name. Why do we need the prefix here?

Wouldn't this break TypeScript use cases where an import for `react` would give you the corresponding types but `thirdparty/react` wouldn't resolve them anymore?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+- **Transitive Dependency Consolidation** — Before Rollup runs, each scanned package's full transitive dependency tree is enumerated by walking `package.json` `dependencies` fields recursively. Any transitive dependency appearing in two or more trees must be externalized — bundled as its own standalone AMD module rather than being inlined into each consumer's bundle. This produces an externals map (which deps to exclude per package) and may add newly-discovered shared packages to the bundle set (see [Section 3.8.3](#383-externals-discovery-via-transitive-dependency-consolidation)).
```
RandomByte, line 72: Keep in mind that `@ui5/project` already contains some functionality to traverse npm dependencies which could be reused here: https://github.com/UI5/cli/blob/main/packages/project/lib/graph/providers/NodePackageDependencies.js

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| Class | Extension | Purpose |
+|-------|-----------|---------|
+| `ModuleInfo` | Add `npmDependencies` Set + `addNpmDependency()` method | Store detected NPM packages per module |
+| `JSModuleAnalyzer` | Add `thirdparty/*` pattern detection in `visit()` cases for AMD deps and ESM imports | Detect NPM refs during existing AST walk |
+| `XMLTemplateAnalyzer` | Add `thirdparty.*` xmlns pattern detection | Detect NPM refs in XML views |
+| `ResourceCollector` | Add `_npmPackages` Set + `getNpmPackages()` aggregation | Collect NPM packages across all resources |
+
+#### 3.4.2 The `thirdparty/*` Convention
+
+The expected namespace convention for NPM packages in UI5 code:
```
RandomByte, line 187: See comment above, why are we introducing this convention?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+Every scenario runs **dual bundling**: the custom Rollup pipeline and `ui5-tooling-modules` execution, producing a side-by-side comparison of output size, build time, and format correctness. This ensures functional equivalence with the established community solution.
+
+#### 3.2.3 Scenario Matrix
+
+| Scenario | Package | Format | What It Validates |
+|----------|---------|--------|-------------------|
+| ESM Integration | `nanoid` | Pure ESM | ESM-only package -> clean AMD
+| CJS Integration | `lodash` | Pure CJS | CJS -> ESM -> AMD via `commonjs` plugin |
+| Web Components | `@ui5/webomponents` | ESM + Custom Elements Manifest | Metadata extraction, UI5 wrapper generation, tag 
+| Complex Package | `sinon` | Mixed deps | Deep dependency trees, tree-shaking effe |
```
RandomByte: >  tree-shaking effe

Incomplete sentence

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+Every scenario runs **dual bundling**: the custom Rollup pipeline and `ui5-tooling-modules` execution, producing a side-by-side comparison of output size, build time, and format correctness. This ensures functional equivalence with the established community solution.
+
+#### 3.2.3 Scenario Matrix
+
+| Scenario | Package | Format | What It Validates |
+|----------|---------|--------|-------------------|
+| ESM Integration | `nanoid` | Pure ESM | ESM-only package -> clean AMD
+| CJS Integration | `lodash` | Pure CJS | CJS -> ESM -> AMD via `commonjs` plugin |
+| Web Components | `@ui5/webomponents` | ESM + Custom Elements Manifest | Metadata extraction, UI5 wrapper generation, tag
```
RandomByte: > Metadata extraction, UI5 wrapper generation, tag 

Incomplete sentence

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+|-----------|-------------|----------|
+| **Package names with `.js`** | e.g., `chart.js`, `easytimer.js` — could be confused with file extensions | Pattern extraction stops at the package name boundary; `.js` is part of the name, not a file extension |
+| **Recursive scanning** | A controller imports another controller which imports an NPM package | ResourceCollector aggregates across all analyzed resources; the NPM set grows as files are processed |
+| **devDependencies filtering** | Packages in `devDependencies` should not be bundled by default | Cross-reference detected packages against `dependencies` in `package.json`; warn if found only in `devDependencies` |
+| **Wildcard ignore patterns** | Test files, spec files should not trigger bundling | Configurable scan include/exclude patterns (default: `webapp/**/*.{js,ts,xml}` excluding `test/**`) |
+| **App-local modules** | Files that look like NPM packages but are local to the app | Only trigger on `thirdparty/*` prefix; app-local modules use relative or `sap/*` paths |
+
+#### 3.4.4 Link with package.json
+
+After scanning produces a list of NPM package names, each name is validated against the project's `package.json`:
```
RandomByte, line 287: I fully agree with this. Any referenced dependency must be declared in the project's package.json. I'm wondering though whether this is also required for `ui5-tooling-modules`. Do you know?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+#### 3.5.2 Plugin Chain
+
+Each plugin transforms code at a specific stage. The order is critical — each depends on transformations from previous plugins.
+
+**Plugin Summary Table:**
+
+| Plugin | Purpose | Source |
+|--------|---------|--------|
+| `@rollup/plugin-node-resolve` | Resolve NPM package paths using package.json fields (exports -> browser -> module -> main) | Rollup ecosystem |
+| `pnpm-resolve` | Follow PNPM symlinks to real file paths | ui5-tooling-modules |
```
RandomByte: Do we really need this plugin? UI5 CLI does not support pnpm officially, yet I believe our module resolution also works there since pnpm acts compatible to standard npm in many ways.

UI5 CLI should remain focused on supporting npm. We should try to not create additional functionality for other package managers.

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| `skip-assets` | Skip CSS/image/font imports that Rollup cannot bundle | ui5-tooling-modules |
```
RandomByte: I don't fully understand the purpose of this plugin. In `ui5-tooling-modules` it seems to skip only CSS files. Here we also mention images and fonts. In general: Is this meant to ignore imports for such files in thirdparty code? Why is this not handled by rollup, or rather: Why do we need to write a plugin for this?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| `@rollup/plugin-commonjs` | Convert CJS (`require`/`module.exports`) to ESM for tree-shaking | Rollup ecosystem |
+| `rollup-plugin-polyfill-node` | Browser polyfills for Node.js built-ins (`path`, `buffer`, `events`, `stream`, `util`) | Community plugin |
```
RandomByte: I thought I rememberd a discussion that we don't want to support Node.js packages for the time being, but I might be wrong about that.

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| `@rollup/plugin-replace` | Static replacement of `process.env.NODE_ENV` -> `"production"` enabling dead-code elimination | Rollup ecosystem |
```
RandomByte: Same as above, depends on whether we want to support Node.js-only packages

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| `dynamic-imports` | Preserve dynamic `import()` calls or convert to Rollup chunks | ui5-tooling-modules |
+| `transform-top-level-this` | Normalize UMD `this` references to `undefined` (ES module semantics) | ui5-tooling-modules |
+| `import-meta` | Transform `import.meta.url` to browser-compatible code | ui5-tooling-modules |
+| `inject-esmodule` | Add `__esModule` flag for CJS/ESM default export interop | ui5-tooling-modules |
+| Rollup AMD output | Built-in: `format: "amd"` with `amd.define: "sap.ui.define"` | Rollup core |
```
RandomByte: We should explain the purpose of these plugins in the concept. Not everything is clear to me and I believe there's room for discussion in what we actually need in UI5 CLI.

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+// ✅ Bundle contains: A + its internal helpers only
+// ❌ Eliminated: B, C, and 297 other unused functions
+// 📦 Example: lodash-es 70KB -> 2KB (97% eliminated)
+```
+
+**When tree-shaking works well:** ESM packages with granular exports (e.g., `lodash-es`).
+
+**When it's limited:** CJS packages where `require()` returns a dynamic object — the `commonjs` plugin attempts to detect named exports but may not achieve full elimination. Packages declaring `"sideEffects": false` in their package.json enable more aggressive elimination.
+
+---
```
RandomByte, line 371: I'm missing some ideas on how bundling and especially tree-shaking could be improved when using TypeScript to write UI5 code. I would expect better results since we can detect which symbols of a package are used in the UI5 project.

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+### 3.6 Build Task Integration
+
+#### 3.6.1 Task Placement
+
+The NPM bundling task integrates into the existing build pipeline as a standard task:
+
+![NPM Build Task Diagram](./resources/0020-npm-integration/build-task-sequence.svg)
+
+**Why after `generateFlexChangesBundle`?** The scanner needs all source resources to be resolved and available. **Why before `generateComponentPreload`?** The generated NPM bundles must be included in the Component-preload.js for production deployment.
```
RandomByte: > The generated NPM bundles must be included in the Component-preload.js for production deployment.

This is an interesting point. We take extra steps to define rollup entry points so we can individual thirdparty-bundles but then end up bundling all of them in the same Component-preload? I'm missing some differentiation for when the standalone thirdparty bundles are relevant at runtime and when they would be consumed through the Component-preload (or library-preload for that matter).

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+#### 3.6.2 Task Interface
+
+The task follows the standard UI5 builder task signature:
+
+- Receives `workspace` (read/write resources), `taskUtil` (project metadata), and `options` (configuration from ui5.yaml)
+- Reads NPM package list from ResourceCollector (populated during earlier analysis phases)
```
RandomByte: There is no earlier analysis phase if the bundling tasks are executed only after this task

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| **Scope** | Global — shared across all apps on the page | Per-app — scoped to the component namespace |
+| **FLP-compatible** | No — collisions when multiple apps ship the same package | Yes — each app's resources are isolated |
+| **Component-preload** | Thirdparty files excluded (different path root) | Thirdparty files included (same namespace root) |
+
+**Why `addToNamespace: true` is the default:** Fiori Launchpad hosts multiple UI5 applications on the same page inside a shell. Each app's resources must live under its own component namespace to avoid collisions. If two apps both ship `thirdparty/lodash.js` at the global path, one overwrites the other — potentially with an incompatible version. With `addToNamespace: true`, each app gets its own copy at its own namespaced path. Apps that don't need FLP compatibility can opt out with `addToNamespace: false` to get shorter paths.
+
+#### 3.6.4 Dependency Rewriting
+
+When `addToNamespace: true`, source file references to NPM packages must be rewritten to include the component namespace prefix. The following types of rewriting occur:
```
RandomByte, line 409: Will this be done as part of the same task as the bundle creation?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+#### 3.6.5 Architectural Decision: Integrated Scanning vs. AMD->ESM Transpilation
+
+An alternative approach was considered: transpile all UI5 AMD source files to ESM format, then let Rollup discover NPM package imports automatically. This was **rejected** for the following reasons:
+
+| Criterion | Integrated Scanning (chosen) | AMD->ESM Transpilation (rejected) |
+|-----------|------------------------------|----------------------------------|
+| **Parse passes** | 1 (existing AST walk) | 3+ (original parse + transpile + Rollup re-parse) |
+| **Temporary files** | None | Must write transpiled ESM to disk, then discard |
+| **Discovery benefit** | Direct extraction | **None** — Rollup still needs explicit entry points; it cannot discover which NPM packages the app uses |
```
RandomByte: But can't we simply provide `Component.js` as the entry point and rollup then discovers all dependencies?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+This rewriting runs as a post-bundling step, after all NPM bundles are generated and placed in the workspace.
+
+#### 3.6.5 Architectural Decision: Integrated Scanning vs. AMD->ESM Transpilation
+
+An alternative approach was considered: transpile all UI5 AMD source files to ESM format, then let Rollup discover NPM package imports automatically. This was **rejected** for the following reasons:
+
+| Criterion | Integrated Scanning (chosen) | AMD->ESM Transpilation (rejected) |
+|-----------|------------------------------|----------------------------------|
+| **Parse passes** | 1 (existing AST walk) | 3+ (original parse + transpile + Rollup re-parse) |
+| **Temporary files** | None | Must write transpiled ESM to disk, then discard |
```
RandomByte: Not sure about this. I believe rollup is capable of working on in-memory files

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| **Discovery benefit** | Direct extraction | **None** — Rollup still needs explicit entry points; it cannot discover which NPM packages the app uses |
+| **XML views** | Pattern-match xmlns attributes directly | Additional transpilation from XML to JS -> ESM |
```
RandomByte: True, although UI5 linter already implements such an XML to ESM transpiler

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+│  ├─ lodash/     (real files)        │  ├─ .pnpm/          <- real files in store
+│  └─ react/      (real files)        │  │  ├─ lodash@4.17.21/node_modules/lodash/
+├─ app1/                              │  │  └─ react@18.3.1/node_modules/react/
+│  └─ node_modules/ (usually empty)   │  ├─ lodash -> .pnpm/lodash@.../  <- symlink
+└─ app2/                              │  └─ react  -> .pnpm/react@.../   <- symlink
+   └─ node_modules/ (usually empty)   └─ app1/
+                                          └─ node_modules/ <- symlinks to root
+```
+
+For PNPM, `preserveSymlinks: false` (the Rollup default) ensures symlinks are followed to the real file paths in the `.pnpm/` store. The `pnpm-resolve` plugin from `ui5-tooling-modules` provides additional handling for edge cases like workspace package.json lookup through symlink chains.
```
RandomByte, line 482: Expand on those additional edge cases if we want to take over that `ui5-tooling-modules` plugin

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+The `nodeResolve` plugin follows this priority when resolving a package:
+
+The `nodeResolve` plugin follows this priority waterfall (see also the resolution cascade in [3.4.1](#341-design-integrated-into-existing-ast-parsing)): **exports** (highest priority, modern packages) -> **browser** (browser-specific overrides) -> **module** (ESM entry, preferred for tree-shaking) -> **main** (CJS entry, fallback) -> **index.js** (convention-based fallback). The exports field supports conditional resolution with `browser` > `import` > `require` > `default` priority.
+
+With `browser: true` configured, the resolution prefers browser-specific builds. This is critical for packages like `axios` that provide separate browser and Node.js entry points.
+
+#### 3.7.3 Advanced Edge Cases
+
+These edge cases are handled based on `ui5-tooling-modules` production experience:
```
RandomByte, line 492: This describes what `ui5-tooling-modules` provides. We should also document which of these we need to implement in UI5 CLI as well

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+| **Workspace dependencies** | `"app2": "workspace:*"` — another UI5 app, not an NPM package | Filter workspace packages by checking root `package.json` workspaces list |
+| **Conditional exports priority** | Browser vs. Node.js conditions in exports field | Use `browser` > `import` > `require` > `default` priority order |
+
+---
+
+### 3.8 Shared Dependencies
+
+#### 3.8.1 Design Principle
+
+**Core concept:** One NPM package = One AMD module file.
```
RandomByte, line 508: I'm confused by this. Won't rollup in some cases bundle multiple npm packages into one AMD module that is then imported in the app?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+| Approach | react-dom | react-colorful | react (shared) | Total Size | React Instances |
+|----------|-----------|----------------|----------------|------------|-----------------|
+| **WITHOUT externals** | 596 KB<br/>(includes React 19KB) | 33 KB<br/>(includes React 19KB) | — | **629 KB** | 2x ❌ Duplicated |
+| **WITH externals** | 577 KB<br/>(references external) | 14 KB<br/>(references external) | 19 KB<br/>(loaded once) | **610 KB** | 1x ✅ Shared |
+| **Savings** | -19 KB | -19 KB | +19 KB | **-19 KB** | Deduplication |
+
+The externals + paths mechanism solves this: `react-dom` and `react-colorful` declare `react` as external, and their AMD output references `thirdparty/react` instead of inlining React's code.
+
+#### 3.8.3 Externals Discovery via Transitive Dependency Consolidation
```
RandomByte, line 531: Is my understanding correct that this will also define the entry points for rollup?

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+- Not all shared dependencies declare peer dependencies (e.g., `scheduler` inside `react-dom`)
+- It detects **actual sharing** based on the concrete dependency graph rather than declared intentions
+- It works identically regardless of whether the shared dep is a peer dependency, a regular dependency, or both
+
+**Relationship with "Let Rollup Handle the Rest" ([Section 3.4.5](#345-let-rollup-handle-the-rest))**: Rollup still handles all file-level resolution (entry points, internal imports, polyfills) during bundling. The consolidation step only determines *which packages* should be external vs. inlined — it reads `package.json` dependency metadata, not source code.
+
+#### 3.8.4 Build Order Optimization
+
+At **build time**, packages must be bundled in topological order (dependencies first) so that externals are resolved correctly:
```
RandomByte, line 576: Is this something we need to be aware of? I would think this is handled within rollup.

rfcs/0020-npm-package-integration.md
```diff
@@ -0,0 +1,807 @@
...
+
+Each patch addresses a specific framework limitation (e.g., boolean attribute rendering, event parameter mapping, value state inconsistencies). As the framework evolves, patches are progressively removed. The patch set is sourced from `ui5-tooling-modules`'s production-tested `rollup-plugin-webcomponents.js`.
+
+---
+
+### 3.10 Package Patching
+
+#### 3.10.1 Motivation
+
+NPM packages sometimes contain bugs, security vulnerabilities, or incompatibilities that the upstream maintainer has not yet fixed — or cannot fix in a timely manner. In these cases, the consuming project must apply a targeted fix (patch) to the package source *before* Rollup bundles it. Without a patching mechanism, the only alternatives are forking the package or waiting for an upstream release.
```
RandomByte, line 722: I think we should discuss whether patching should be part of the initial release of this feature. I'm still uncomfortable adding such functionality while most other web frameworks do not provide something similar.

rfcs/0020-npm-package-integration.md
d3xter666: Couple of things:
- Extraction of shared packages into standalone packages
- This dedublication will lead to smaller parent packages, less size being downloaded, browser caching

rfcs/0020-npm-package-integration.md
d3xter666: This is an interesting point and I fully agree with it, but yes, it's kind of "new" concept here.

Here are my toughts on this:
- If we have a direct package name being referenced directlty i.e. `react`, this means that this package is not UI5 AMD wrapped and the code that follows knows how to handle the raw format. The package is taken directly from `node_mdules/`
- Our current situation (in UI5 repositories) is that all the packages/thirdparties that are (usually) UI5 AMD wrapped and stay in a corresponding `thirdparty/` folder.

rfcs/0020-npm-package-integration.md
d3xter666: Thanks a lot!
I'll take a look at it. As you can see in the PoC, currently we have the "external" (transitive) dependencies being hardocded as I wanted to try the feasibility of the build and the expected output

rfcs/0020-npm-package-integration.md
d3xter666: It would be a lot easier if we have a concept like this one.
Otherwise, we must do some kind of heuristics in order to identify external packages which might be fragile and hard to implement.
https://github.com/UI5/cli/pull/1287#discussion_r2958451270

rfcs/0020-npm-package-integration.md
d3xter666: This is the biggest gap between `ui5-tooling-modules` and the current PoC.
The `ui5-tooling-modules` tries to fullfill the current situation and this requirement is not a must.
It has a heavy heuristics to identify modules/packages.

Having 3rd parties defined as dependencies in `package.json` solves many of edge cases `ui5-tooling-modules` tries to resolve.

We must stick to that requirement in order to provide a clean and robust solution.
The idea for `thirdparties/` "heuristics" build on top of that: https://github.com/UI5/cli/pull/1287#discussion_r2953287805

rfcs/0020-npm-package-integration.md
d3xter666: The idea of this plugin is to give you the currect location to ceratin module. As far as I got it, `pnpm` creates symlinks in order to optimise for space and size. This might result in rollup not being able to resolve correctly physical phats to certain dependencies i.e. this plugin might not work correctly: https://www.npmjs.com/package/@rollup/plugin-node-resolve

Basically, we do not modify the package or something, but provide hints to rollup in certain environments to provide the correct output regardless of the workspace setup

rfcs/0020-npm-package-integration.md
d3xter666: It is impossible to support "backend" packages on the frontend completelly.
This is a knowledge gained from `ui5-tooling-modules`.
I tried to cover and list all the possible cases I was able to identify. Of course, we can skip certain scenarios

rfcs/0020-npm-package-integration.md
d3xter666: https://github.com/UI5/cli/pull/1287#discussion_r2958539057

rfcs/0020-npm-package-integration.md
d3xter666: As I mentioned above, these are rollup plugins used in `ui5-tooling-modules`. Some of them cover certain edge cases that I didn't go deep into, others are kind of stadartized rollup hints i.e. `Add "__esModule" flag...`

I will provide more detailed explanation of those plugins and their reasoning

rfcs/0020-npm-package-integration.md
d3xter666: I didn't cover this case for several reasons:
- TypeScript uses ESM which is native to Rollup
- When the transpilation from TypeScript to JS happens in the tooling?

rfcs/0020-npm-package-integration.md
d3xter666: Maybe this is a miss from my side!
And now when you mention it, makes sense.

I am not sure how this works currently with thirdparties in UI5. Are those bundled or are requested separately?

rfcs/0020-npm-package-integration.md
d3xter666: That's exactly the idea- to discuss what we can and want to deliver

rfcs/0020-npm-package-integration.md
d3xter666: Maybe, I need to be more clear here. It's more like:
one entry point (import/require/sap.ui.define) in UI5 source files === npm package
But this does not mean that the npm package is standalone. The build is handled by rollup and It can be bundled along with its transitive dependencies.

rfcs/0020-npm-package-integration.md
d3xter666: Yes!
This is the idea of deduplication of (transitive dependencies) packages that will be resued by other packages. 
Those transitive dependencies won't be used directly in UI5's app source code, but by thirdparties, dependending on them

rfcs/0020-npm-package-integration.md
d3xter666: Unfortunatelly, not!
As far as I got this, we must first build the "clean" dependencies, so that they can be reused (as imports) for the rest of bundling. Otherwise, rollup might:
- not discover them correctly
- bundle everything

rfcs/0020-npm-package-integration.md
d3xter666: We must definitelly discuss this! It's just an idea as we have a lot of packges being patched currently.
However, when we have the possibility to update and monitor packages quickly via `package.json`, the idea of patching might become redundant.

WebComponents currently do that, for example: https://github.com/UI5/webcomponents/tree/main/.yarn/patches
They rely on yarn, though!

rfcs/0020-npm-package-integration.md
d3xter666: I think it highly depends on "who" will use the bundling.
Yes, this is an extraction from `ui5-tooling-modules`, but I believe it was done there, because someone used `pnpm` in their repositories and it fell into certain issue.

As we discussed earlier, I wanted to summarize all the information I was able to collect for this topic.
It's up to us and the scenarios we want to addrees, what we will implement or not.

rfcs/0020-npm-package-integration.md
d3xter666: Thanks!

rfcs/0020-npm-package-integration.md
d3xter666: Thanks!

rfcs/0020-npm-package-integration.md
d3xter666: Some NPM packages contain import './styles.css' or similar non-JS import statements in their source code. When Rollup encounters these during bundling, it fails because it doesn't know how to handle CSS files — they're not JavaScript modules.

I will add a more detailed explanations for those plugins. Still, they are used within the `ui5-tooling-modules` and I wanted to be sure that we do not miss something

rfcs/0020-npm-package-integration.md
d3xter666: Maybe, this is a redundant step and we can consolidate this during the build and write

rfcs/0020-npm-package-integration.md
d3xter666: I have rewritten this section with a bit more details

rfcs/0020-npm-package-integration.md
d3xter666: I have updated the document

rfcs/0020-npm-package-integration.md
d3xter666: I have refactored the document to try to explain my points clearly

rfcs/0020-npm-package-integration.md
d3xter666: My hope was the same initially and along with the idea to transpile all the UI5 Code to ESM, this seemed like the perfect fit!

However, this is not the case! If we feed the Component.(js|ts) directly to the rollup (assuming somehow rollup knows about module resolution in UI5), it will create a single bundle with everything inside.

In order to have thirdparties as stadlone builds, we must:
1. Identify those thirparty components
2. Build them standalone

rfcs/0020-npm-package-integration.md
d3xter666: Thanks! I have refactored this

rfcs/0020-npm-package-integration.md
RandomByte: I believe they are bundled by default but we manually exclude some. In sap.ui.core we have a custom bundle with selected thirdparties: https://github.com/UI5/openui5/blob/master/src/sap.ui.core/ui5.yaml#L43-L51

So I guess it depends on the project's individual use case for each thirdparty.

rfcs/0020-npm-package-integration.md
RandomByte: Do we have an example for such an npm package? If rollup does not handle that, I'm wondering what tool they use.

d3xter666: webcomponents, for example

PR #1294, opened 2026-02-06
ci: Replace depcheck with knip

werwolf2303:
Due to the deprecation of depcheck we migrated to knip, which is in active development and has more features. As a first step we only activate the checking of unused dependencies except peer dependencies (See comment in config). In the future we may also activate more rules to improve code quality.

For monorepos it has to be run at root package level: [Knip Monorepos](https://knip.dev/features/monorepos-and-workspaces).

The reason why we replaced the `docdash` jsdoc template path lookup in the jsdoc script is that knip fails to detect that a dependency (`jsdoc`) is used when a command inside of the package.json contains JavaScript code (via sub-shell).

Also removes unused local-web-server dependency.

JIRA: CPOUI5FOUNDATION-1126


packages/logger/package.json
d3xter666, line 1: This file is marked as deleted and the CI run states it's missing it

internal/benchmark/package.json
```diff
@@ -11,18 +11,17 @@
...
 	},
 	"scripts": {
-		"test": "npm run lint && npm run coverage && npm run depcheck",
+		"test": "npm run lint && npm run coverage",
 		"unit": "node --test 'test/e2e/**/*.js'",
 		"unit-watch": "node --test --watch 'test/e2e/**/*.js'",
 		"coverage": "node --test --experimental-test-coverage 'test/e2e/**/*.js'",
 		"lint": "eslint .",
-		"depcheck": "depcheck"
+		"depcheck": ""
```
d3xter666: Can we run the `depcheck` script separately for each package?
This way we would have a finer control over the packages.

With the current solution, we have visbility only over the whole monorepository.

Take for example the ESLint check- we have a general setup for all of the packages and then this configuration gets extended within the packages and is run independently in there

internal/shrinkwrap-extractor/package.json
```diff
@@ -19,12 +19,12 @@
 		"directory": "internal/shrinkwrap-extractor"
 	},
 	"scripts": {
-		"test": "npm run lint && npm run coverage && npm run depcheck",
```
matz3, line 22: Please keep the script execution in the `test` script. The idea of `test` is to cover all checks of a package in one script. This is currently only used locally, as we use separate steps (with individual `npm run` commands) in our GitHub Actions CI.

_(same comment for all other occurrences of this removal)_

packages/cli/knip.config.js
```diff
@@ -0,0 +1,15 @@
+import commonConfig from "../../knip.common.config.js";
+
+commonConfig.ignoreDependencies.push(
+	"rimraf",
+	"@istanbuljs/esm-loader-hook",
+	"data-with-position",
```
matz3: Have you checked why some of the "normal" dependencies still need to be listed here?
For some, like `@istanbuljs/esm-loader-hook` I understand that knip does not parse all npm scripts to find the reference of loaders. But for those, like `js-yaml` and `data-with-position` I would have hoped that knip is able to do better than depcheck. Now it seems like we even need more ignores than before, which isn't what I would have expected.

package.json
```diff
@@ -23,12 +23,11 @@
 		"npm": ">= 8"
 	},
 	"scripts": {
-		"test": "npm run lint && npm run check-licenses && npm run depcheck && npm run coverage && npm run schema-generate && npm run generate-cli-doc",
+		"test": "npm run lint && npm run check-licenses && npm run knip --workspaces && npm run coverage && npm run schema-generate && npm run generate-cli-doc",
```
matz3: Please also add a `knip` script, as a replacement for `depcheck` in this package (to run the tool on all packages). It can then be invoked here in the `test` script.

internal/benchmark/knip.config.js
```diff
@@ -0,0 +1,7 @@
+import commonConfig from "../../knip.common.config.js";
+
+commonConfig.ignoreDependencies.push(
+	"eslint"
```
matz3: I would not expect that we need to ignore eslint, as it is used. Maybe the plugin needs to be configured differently? https://knip.dev/reference/plugins/eslint
In our setup we have also a `eslint.common.config.js`, which doesn't match the default patterns.

In case you already checked this, please add a comment that describes why it still needs to be ignored here.

packages/builder/knip.config.js
```diff
@@ -0,0 +1,11 @@
+import commonConfig from "../../knip.common.config.js";
+
+commonConfig.ignoreDependencies.push(
+	"@istanbuljs/esm-loader-hook",
+	"rimraf",
+	"cross-env",
+	"eslint",
+	"nyc"
```
matz3: Similar as for ESLint, there is a plugin that should cover nyc: https://knip.dev/reference/plugins/nyc

knip.common.config.js
```diff
@@ -0,0 +1,23 @@
+const config = {
```
matz3: Which config is used now? This one or the knip.config.json?
I like the usage of a JSON schema for validation, but it prevents us from adding comments, right? For some configuration, a comment would be helpful to express why it was configured that way.

internal/benchmark/package.json
```diff
@@ -11,18 +11,18 @@
...
 	},
 	"scripts": {
-		"test": "npm run lint && npm run coverage && npm run depcheck",
+		"test": "npm run lint && npm run coverage",
 		"unit": "node --test 'test/e2e/**/*.js'",
 		"unit-watch": "node --test --watch 'test/e2e/**/*.js'",
 		"coverage": "node --test --experimental-test-coverage 'test/e2e/**/*.js'",
 		"lint": "eslint .",
-		"depcheck": "depcheck"
+		"knip": "knip --config knip.config.js"
```
matz3: This script is not needed anymore, isn't it?

internal/benchmark/package.json
```diff
@@ -11,18 +11,18 @@
...
 	},
 	"dependencies": {
 		"js-yaml": "^4.1.1"
 	},
 	"devDependencies": {
-		"depcheck": "^1.4.7",
-		"eslint": "^9.39.1"
+		"eslint": "^9.39.1",
+		"knip": "^5.83.1"
```
matz3: Dependency not needed here

internal/documentation/package.json
```diff
@@ -24,14 +24,14 @@
 	"scripts": {
 		"start": "vitepress dev --open",
 		"lint": "eslint .",
-		"depcheck": "depcheck --ignores @ui5/documentation,vitepress,docdash,jsdoc,open-cli,handlebars,@types/node,@ui5/webcomponents,autoprefixer,cssnano,@theme/components",
 		"dev": "vitepress dev",
 		"build:vitepress": "vitepress build",
 		"build:assets": "sh -c 'DEST_DIR=${1:-./dist}; cp -r ./docs/images \"$DEST_DIR/images\"' --",
 		"preview": "vitepress preview --port 8080",
+		"prepare-docdash": "./scripts/prepareDocdash.sh",
```
matz3: I don't think this script is related to the current PR / change.

packages/server/package.json
```diff
@@ -90,7 +89,7 @@
 		"@ui5/builder": "^5.0.0-alpha.2",
 		"@ui5/fs": "^5.0.0-alpha.2",
 		"@ui5/logger": "^5.0.0-alpha.2",
-		"body-parser": "^2.2.2",
+		"body-parser": "^1.20.2",
```
matz3: Why is this package downgraded?

knip.config.json
```diff
@@ -0,0 +1,48 @@
+{
```
matz3: See https://github.com/UI5/cli/pull/1294#discussion_r2792062048.

The comments from the config are gone now. I would prefer to have at least some kind of comment on why we are not activating the rules. A short summary of the purpose of this change in the commit message / PR description would be helpful (in short: This change is supposed to be a replacement for depcheck, so only those checks/rules are used, but in future also more checks of knip can be enabled).
This will help everyone, especially people who were not initially involved, to understand the reasons behind the changes and configuration.

In addition, it would be helpful to also add comments to the ignores to explain why they are needed.
For this reason, I would propose to rename the config to  `knip.config.jsonc`, to be able to add comments, but still benefit from the JSON schema for validation. This seems to be supported by knip.

knip.config.json
```diff
@@ -0,0 +1,48 @@
...
+		"@istanbuljs/esm-loader-hook"
+	],
+	"workspaces": {
+		".": {
+			"ignoreDependencies": [
+				"husky"
+			]
+		},
+		"packages/cli": {
+			"entry": ["lib/framework/*.js", "lib/cli/commands/*.js"]
```
matz3: I would assume that the `framework` files don't need to be listed explicitly here. Only the `commands` modules are dynamically imported, which requires this additional configuration.

knip.config.json
```diff
@@ -0,0 +1,48 @@
...
+		"catalog": "off",
+		"exports": "off",
+		"types": "off",
+		"enumMembers": "off",
+		"optionalPeerDependencies": "off"
+	},
+	"ignoreDependencies": [
+		"@ui5/webcomponents",
+		"@ui5/fs",
+		"@ui5/server",
```
matz3: I get the following output:
```
Configuration hints (1)
@ui5/server    knip.config.json  Remove from ignoreDependencies
```

Is this ignore entry really needed?

knip.config.json
```diff
@@ -0,0 +1,48 @@
...
+		"binaries": "off",
+		"unresolved": "off",
+		"catalog": "off",
+		"exports": "off",
+		"types": "off",
+		"enumMembers": "off",
+		"optionalPeerDependencies": "off"
+	},
+	"ignoreDependencies": [
+		"@ui5/webcomponents",
```
matz3: @d3xter666 I can't find any usage of "@ui5/webcomponents" in the Vitepress setup. Do you know where this is used / needed? If it's not needed, we should remove the dependency rather than ignoring it here.

knip.config.json
```diff
@@ -0,0 +1,48 @@
...
+		"@ui5/fs",
```
matz3: `@ui5/fs` is only reported for `@ui5/cli`, so should we rather limit the ignore to just the cli package?
The reason for the dependency is that we want to ensure the expected versions of all packages when using the CLI. But the `@ui5/fs` package is not used / imported directly from CLI code, except for a generic require of the package.json (see `packages/cli/lib/cli/commands/versions.js`).

knip.config.json
```diff
@@ -0,0 +1,48 @@
...
+				"husky"
+			]
+		},
+		"packages/cli": {
+			"entry": ["lib/framework/*.js", "lib/cli/commands/*.js"]
+		},
+		"packages/server": {
+			"ignoreDependencies": [
+				"compression",
+				"cors"
```
matz3: Those are dynamic imports via `packages/server/lib/middleware/MiddlewareManager.js` (should be mentioned in a comment in the config)

internal/documentation/package.json
```diff
@@ -24,22 +24,20 @@
...
 		"jsdoc": "npm run jsdoc-generate && open-cli dist/api/index.html",
-		"jsdoc-generate": "jsdoc -c jsdoc/jsdoc-workspace.json -t $(node -p 'path.dirname(require.resolve(\"docdash\"))') ./ || (echo 'Error during JSDoc generation! Check log.' && exit 1)",
-		"jsdoc-generate-gh-pages": "jsdoc -c jsdoc/jsdoc.json -t $(node -p 'path.dirname(require.resolve(\"docdash\"))') ./ || (echo 'Error during JSDoc generation! Check log.' && exit 1)",
-		"generate-cli-doc": "node ./scripts/generateCliDoc.js",
```
matz3, line 35: This script should not have been removed, isn't it?
When adding back the script, the configuration for `entry: ["scripts/*.js"]` is not needed anymore as the script usage is detected.

werwolf2303: I think that happened when I reset the head

knip.config.json
d3xter666: It seems to be a leftover from the initial implementation. 
Just checked the final build of vitepress docs and there are no `<ui5-` tags in there. 
I was thinking also that it could be the theming, but it seems to be fetched from the CDN.
So, I think, it's ok to be removed

d3xter666 (commented):
LGTM

knip.config.json
werwolf2303: Ok. I will keep the change

PR #1299, opened 2026-02-09
ci(documentation): Improve integrity of documentation build

d3xter666:
Currently, if there's a package update of ceratin documentation dependencies, we cannot be sure whether this can break the vitepress build. 

This integrity test checks exactly this.

An example is https://github.com/UI5/cli/pull/1295


matz3:
The commit message was changed, but the PR was still merged with the previous PR title (invalid according to commitlint).
I think we should consider using a merge queue with the commitlint check to prevent this. It already happened several times in the past.

PR #1303, opened 2026-02-13
ci: Fix Vitepress integrity check

d3xter666:


matz3:
Could you please shortly explain where this change is relevant or where something went wrong to cause problems?

PR #1312, opened 2026-02-23
build(deps-dev): Bump minimatch  to 10.2.2

d3xter666:
Fixes: https://github.com/advisories/GHSA-3ppc-4f35-3m26

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -94,9 +94,9 @@
 			}
 		},
 		"internal/shrinkwrap-extractor/node_modules/@npmcli/arborist": {
-			"version": "9.3.0",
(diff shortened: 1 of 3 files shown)


maxreichmann:
What is our approach regarding dependencies using minimatch prior to version 10? E.g. in the builder package, ava is depending on minimatch@9.0.5.
The vulnerability report states all versions prior to 10.2.1 are affected. I guess, we can only update version 10?

d3xter666:
> What is our approach regarding dependencies using minimatch prior to version 10? E.g. in the builder package, ava is depending on minimatch@9.0.5. The vulnerability report states all versions prior to 10.2.1 are affected. I guess, we can only update version 10?

This change will only update production dependencies! Unfortunatelly at this point we cannot do much for those `eslint` deps

PR #1329, opened 2026-03-20
fix(builder): JSDoc: avoid errors when symbol has no events

maxreichmann:
This is a fix for a bug which initially has been fixed in v4 (https://github.com/SAP/ui5-builder/pull/1162) but not upported to v5 yet. This fix also resolves the error occurrence found when running `ui5 build jsdoc` in the UI5/sample-app.
~~This will add a unit test for transformApiJson.js covering this scenario.~~

### DIFF
--- a/packages/builder/lib/processors/jsdoc/lib/transformApiJson.js
+++ b/packages/builder/lib/processors/jsdoc/lib/transformApiJson.js
@@ -1034,15 +1034,41 @@ function transformer(sInputFile, sOutputFile, sLibraryFile, vDependencyAPIFiles,
 			const componentDir = path.join(sSectionsDir, sComponentPath);
 			if (fs.existsSync(componentDir)) {
 				try {
+				const dirContents = fs.readdirSync(componentDir, { withFileTypes: true });
 
-					const customSections = fs.readdirSync(componentDir, { withFileTypes: true })
-						.filter((dirent) => dirent.isFile() && dirent.name.endsWith('.md'))
-						.map((dirent) => dirent.name.replace(/\.md$/, ''));
-
-					// Store list of available sections for this symbol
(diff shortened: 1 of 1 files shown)


packages/builder/lib/processors/jsdoc/lib/transformApiJson.js
RandomByte: Maybe let's just simplify this:

```suggestion
								const oParamInSymbol = oEvtInSymbol?.parameters?.[0]?.parameterProperties?.getParameters?.parameterProperties?.[sParam];
```

codeworrior:
While I like the idea of unit tests in general,  I think it is a bad idea to have this test only in the @ui5/cli. The transformApiJson is mainly developed in openui5/lib/jsdoc. If the test does not exist there, risk is high that the next sync from openui5 to @ui5/cli gets unnecessarily difficult. I would in general prefer to make openui5/lib/jsdoc the single source of truth and the copy here just a copy.

There is even a test for the transformer, but it's located in a different place (for historical reasons, I guess)
- sapui5.runtime/src/dist/docbuilder-maven-plugin/src/test/tools/node/test/test-transformer.js

packages/builder/lib/processors/jsdoc/lib/transformApiJson.js
codeworrior: Again, I would suggest NOT to do this refactoring., or at least not in this repo first, as it unnecessarily complicates the cherry picks from SAP/openui5 lib/jsdoc.

I've reverted this and instead pushed all missing changes from SAP/openui5 lib/jsdoc.

maxreichmann:
Just for documentation purposes (an earlier comment with the same message is gone):
This PR has been updated with all missing updates from the original source: openui5/lib/jsdoc

PR #1354, opened 2026-04-07
fix(server): Explicitly bind to IPv4 loopback

RandomByte:
In some environments, 'localhost' might resolve to the IPv6 loopback address '::1' which is often unexpected. This can especially lead to problems where other tools might resolve 'localhost' differently in the same environment.

This change explicitly binds to the IPv4 loopback address '127.0.0.1' instead of resolving the address from 'localhost'.

In case remote connections are allowed, no host is specified so that Node.js binds to either the unspecified IPv4 or -IPv6 address (which usually also accepts IPv4 connections) [1]

[1] https://nodejs.org/docs/latest-v25.x/api/net.html#serverlistenport-host-backlog-callback

Cherry-pick of https://github.com/SAP/ui5-server/pull/763

### DIFF
--- a/packages/server/lib/server.js
+++ b/packages/server/lib/server.js
@@ -26,10 +26,11 @@ function _listen(app, port, changePortIfInUse, acceptRemoteConnections) {
 		const options = {};
 
 		if (!acceptRemoteConnections) {
-			options.host = "localhost";
-		}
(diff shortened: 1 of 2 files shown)


RandomByte:
Note that this is an up-port of https://github.com/SAP/ui5-server/pull/763

d3xter666 (approved):

PR #1355, opened 2026-04-08
test: Add e2e-tests

maxreichmann:
Outsourced of [Incremental Build PR #1267](https://github.com/UI5/cli/pull/1267):

* Add "e2e-tests" dir under "internal/"
* Add tests for following community custom tasks:
  * ui5-tooling-transpile
  * ui5-task-zipper
  * ui5-tooling-modules
  * ui5-tooling-stringreplace
* Add test for "ui5 --version"

All tests are currently passing (April 8th).


maxreichmann:
@UI5/ui5-team-cor-fnd These tests are not included in any CI pipeline yet. It's up for debate if we want to include them.

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,227 @@
...
+		this.fixtureName = fixtureName;
+		this.originFixturePath = path.resolve(__dirname, "../fixtures", fixtureName);
+		this.tmpPath = path.resolve(__dirname, "../tmp", fixtureName);
+		this.dotUi5Path = path.resolve(this.tmpPath, ".ui5");
+		this.distPath = path.resolve(this.tmpPath, "dist");
+	}
+
+	async init() {
+		// Clean up previous runs
+		await fs.rm(this.tmpPath, {recursive: true, force: true});
```
d3xter666, line 28: Why is the `tmp` path part of the commit, then?

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,227 @@
...
+			});
+		});
+	}
+}
+
+describe("ui5 build", () => {
+	test("ui5-tooling-transpile", async ({assert}) => {
+		const fixtureHelper = new FixtureHelper("application.a.ts");
+		await fixtureHelper.init();
+		process.env.UI5_DATA_DIR = `${fixtureHelper.dotUi5Path}`;
```
d3xter666: This is manipulated globally and is not cleaned up.
A cleanup in before/after pahse is needed.

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,227 @@
...
+		process.chdir(fixtureHelper.tmpPath);
```
d3xter666: This changes the directory and is not reset later

d3xter666 (changes requested):
There's also an e2e test setup in `packages/cli`.
Can we leverage and consolidate the setup from there?
Or simply clean it up if it's redundant an too complex anymore?

internal/e2e-tests/test/build.js
maxreichmann: Thank you for catching that 👍 Must have been an oversight during the outsourcing process. Fixed now

maxreichmann:
Tests are now included in Github Actions via a separate workflow (`internal-e2e-tests.yml`) and take about a minute of execution.

internal/e2e-tests/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Fixtures
+
+Located under `./fixtures/`:
+
+| Fixture | Description |
+|---|---|
+| `application.a` | Sample JavaScript project (controller + `manifest.json`) |
+| `application.a.ts` | Sample TypeScript project (based on [generator-ui5-ts-app](https://github.com/ui5-community/generator-ui5-ts-app)) |
+
+> **Note:** These tests are not yet included in any CI pipeline.
```
d3xter666: This is not correct anymore. We have a pipeline running for pull request and merge in `main`

d3xter666 (commented):
LGTM

internal/e2e-tests/test/version.js
```diff
@@ -0,0 +1,31 @@
...
+import {fileURLToPath} from "node:url";
+import path from "node:path";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+describe("ui5 version", () => {
+	test("output the version of the UI5 CLI", ({assert}) => {
+		const ui5Path = path.resolve(__dirname, "../../../packages/cli/bin/ui5.cjs");
+		execFile("node", [ui5Path, "--version"], (error, stdout, stderr) => {
```
RandomByte: I think this test is identical to https://github.com/UI5/cli/blob/40afbdf0d47885659cba14830e8501ddcb4eb80a/packages/cli/test/lib/cli/base.js#L40

Do we need both?

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,228 @@
...
+
+	async init() {
+		// Clean up previous runs
+		await fs.rm(this.tmpPath, {recursive: true, force: true});
+		// Copy source files to temp location
+		await fs.cp(this.originFixturePath, this.tmpPath, {recursive: true});
+		// Install node_modules
+		await this._installNodeModules();
+		// Setup environment
+		process.env.UI5_DATA_DIR = `${this.dotUi5Path}`;
```
RandomByte:
```suggestion
		process.env.UI5_DATA_DIR = this.dotUi5Path;
```

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,228 @@
...
+		const componentPreload = await fs.readFile(path.resolve(fixtureHelper.distPath, "Component-preload.js"), "utf-8");
+		assert.ok(componentPreload.includes("application/a/ts/controller/Test.controller"), "Component-preload.js should contain the TS resource transpiled to JS");
+		assert.ok(!componentPreload.includes("randomTSType"), "Component-preload.js should NOT contain any TS syntax");
+		const componentPreloadMap = await fs.readFile(path.resolve(fixtureHelper.distPath, "Component-preload.js.map"), "utf-8");
+		assert.ok(componentPreloadMap.includes("randomTSType"), "Component-preload.js.map should contain the TS type information");
+
+		// --------------------------------------------------------------------------------------------
+
+		// Modify source files
+		await fixtureHelper.prepareForNextRun();
```
RandomByte: Why is this (deletion and fresh copy of sources) necessary here? From what I can see, the source files have not been changed?

internal/e2e-tests/fixtures/application.a.ts/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
RandomByte, line 1: We should have a package-lock.json next to every package.json. Then we can also use ranges for the dependencies

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,228 @@
...
+			console.log(chartJS);
+		}
+	});
+});`;
+		await fs.writeFile(newControllerPath, newControllerContent, "utf-8");
+
+		// #2 Build
+		await fixtureHelper.build(assert, ui5YamlName);
+
+		// Test: the dist contains the new controller and the third party import
```
RandomByte, line 150: Can we also check whether the thirdparty itself is present in the build result? (unless it's already without the import)

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,228 @@
...
+
+
+		// Add a new source file with a placeholder string
+		await fixtureHelper.prepareForNextRun();
+		const newControllerPath = path.resolve(fixtureHelper.tmpPath, "webapp/controller/New.controller.js");
+		const newControllerContent =
+`sap.ui.define([], () => {
+	return Controller.extend("application.a.controller.New",{
+		onInit() {
+			console.log(\${PLACEHOLDER_TEXT});
```
RandomByte, line 179: I think this is not a good example because the result is not valid JavaScript:

```js
console.log(\"INSERTED_TEXT\")
```

internal/e2e-tests/package.json
```diff
@@ -0,0 +1,17 @@
...
+	"engines": {
+		"node": "^22.20.0 || >=24.0.0",
+		"npm": ">= 8"
+	},
+	"scripts": {
+		"test": "node --test 'test/**/*.js'",
+		"test-watch": "node --test --watch 'test/**/*.js'"
+	},
+	"dependencies": {
+		"adm-zip": "0.5.17"
```
RandomByte: Dependency is locked via the root package-lock.json, so we can use a range here:

```suggestion
		"adm-zip": "^0.5.17"
```

internal/e2e-tests/fixtures/application.a.ts/package.json
```diff
@@ -0,0 +1,10 @@
...
+	"name": "application.a.ts",
```
RandomByte: Use a prefix like `@ui5-internal` for test packages

internal/e2e-tests/fixtures/application.a.ts/webapp/controller/Test.controller.ts
```diff
@@ -0,0 +1,22 @@
+type randomTSType = {
+	first: {
+		a: number,
+		b: number,
+		c: number
+	},
+	second: string
+}
+
+export default class Main {
```
RandomByte: Should we use a proper UI5 controller that extends `sap/ui/core/mvc/Controller` like in https://github.com/SAP-samples/ui5-typescript-tutorial/blob/main/exercises/ex4/com.myorg.myapp/webapp/controller/BaseController.ts#L13 ?

RandomByte (changes requested):
Sorry for the late review

internal/e2e-tests/test/version.js
maxreichmann: Removed.

internal/e2e-tests/test/build.js
maxreichmann: Just ran the test and checked the dist output of **controller/New.controller.js**:
`sap.ui.define([],()=>Controller.extend("application.a.controller.New",{onInit(){console.log("INSERTED_TEXT")}}));
//# sourceMappingURL=New.controller.js.map`

Looks fine to me. Or do you mean something different?

RandomByte: My bad, makes sense 👍🏼

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,211 @@
...
+					return;
+				}
+				resolve();
+			});
+		});
+	}
+
+	async _installNodeModules() {
+		await new Promise((resolve, reject) => {
+			execFile("npm", ["install"], { cwd: this.tmpPath }, (error, stdout, stderr) => {
```
RandomByte:
```suggestion
			execFile("npm", ["ci"], { cwd: this.tmpPath }, (error, stdout, stderr) => {
```

Let's use `npm ci` to make sure that the lockfile is being used (the command will fail if no lockfile is present or the lockfile is inconsistent with the package.json).

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,211 @@
...
+	async init() {
+		// Clean up previous runs
+		await fs.rm(this.tmpPath, {recursive: true, force: true});
+		// Copy source files to temp location
+		await fs.cp(this.originFixturePath, this.tmpPath, {recursive: true});
+		// Install node_modules
+		await this._installNodeModules();
+		// Setup environment
+		process.env.UI5_DATA_DIR = this.dotUi5Path;
+		process.chdir(this.tmpPath);
```
RandomByte: Refrain from manipulating `process.chdir`. Rather provide the `cwd` to the `execFile` command (as already done in `_installNodeModules`) as a best practice.

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	async build(assert, ui5YamlName) {
+		await new Promise((resolve, reject) => {
+			execFile("node", [ui5CliPath, "build", "--config", ui5YamlName, "--dest", this.distPath], async (error, stdout, stderr) => {
```
RandomByte: In our other tests we use [execa](https://github.com/sindresorhus/execa) instead of the native Node API (e.g. https://github.com/UI5/cli/blob/40afbdf0d47885659cba14830e8501ddcb4eb80a/packages/cli/test/lib/cli/base.js#L15). Usage is simpler since you get a promise directly, which would save us a bunch of lines here

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,211 @@
+// Tests running "ui5 build"
+// with fixtures (under ../fixtures)
+// and by using node's child_process module and the ui5.cjs under ../../../packages/cli/bin/ui5.cjs.
+
+import { execFile } from "node:child_process";
```
RandomByte: Is ESLint applied to this code? Looks to me like the extra spaces should raise a linting error

internal/e2e-tests/test/build.js
```diff
@@ -0,0 +1,211 @@
...
+
+	async init() {
+		// Clean up previous runs
+		await fs.rm(this.tmpPath, {recursive: true, force: true});
+		// Copy source files to temp location
+		await fs.cp(this.originFixturePath, this.tmpPath, {recursive: true});
+		// Install node_modules
+		await this._installNodeModules();
+		// Setup environment
+		process.env.UI5_DATA_DIR = this.dotUi5Path;
```
RandomByte: Have you tested whether this actually works? In my test I couldn't see a `.ui5` directory inside `tmp`.

This only manipulate the env variable of the current Node.js process (executing the test script). I don't think it's inherited to any processes spawned from here automatically. You rather need to specify this as part of the process spawn command (in the `build` function`)

RandomByte (changes requested):

internal/e2e-tests/test/build.js
maxreichmann: I've just tested it with "application.a.ts" and a .ui5 directory gets successfully generated inside it.
For "application.a" no .ui5 directory gets generated anyway because no extra framework libs need to be saved.

You can actually just try this out with running a normal "ui5 build" inside it and see that under ~/ no ".ui5" dir gets created (assuming you just cleared it beforehand).

This will change once we're bringing Incremental Build (cache) functionality into it.

RandomByte: Alright, great 👌

RandomByte (approved):
Looking good now, thanks for the quick iterations 🙌

d3xter666 (approved):

PR #1356, opened 2026-04-09
ci: Add engine compatibility check for PRs

d3xter666:
Reusable workflow that runs `npm ci --dry-run` against boundary Node versions and posts a consolidated PR comment with incompatible packages. Adds `blocked` label on failure with detailed Github Bot comment with the failing packages and their node engines.

Approaches evaluated:

- `npm ci --engine-strict`: runtime check only, stops at first failure, misses remaining incompatible packages
- `check-engine` (npm): only checks running Node against root engines, doesn't inspect dependency tree
- `check-engine-light` (npm): range-vs-range via semver.subset(), but v0.x, only 4 releases, unstable
- `ls-engines` (npm): range-vs-range capable, but v0.x, 17-month release gap, 7 prod dependencies

**Selected approach:** single-job nvm loop using `npm ci --dry-run` (without `--engine-strict`). Reports all EBADENGINE warnings as a table in one PR comment. No extra dependencies, no matrix, no artifacts — just shell and tools already on the runner. 
**Note:** NVM seems to come bundled in the GitHub action images

PoC of the implementation: https://github.com/d3xter666/test-ci/pull/1
JIRA: CPOUI5FOUNDATION-1225

### DIFF
--- a/.github/workflows/engine-compat.yml
+++ b/.github/workflows/engine-compat.yml
@@ -25,6 +25,7 @@ jobs:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         NODE_VERSIONS: ${{ inputs.node-versions }}
       run: |
+        # Load nvm (pre-installed on GitHub runners)
         export NVM_DIR="$HOME/.nvm" && . "$NVM_DIR/nvm.sh"
 
         PR="${{ github.event.pull_request.number }}"
@@ -32,6 +33,7 @@ jobs:
         FAILED=false
         TABLE="| Node | Result | Incompatible Packages |\n|---|---|---|"
 
+        # Loop through each requested Node version
         IFS=',' read -ra VERSIONS <<< "$NODE_VERSIONS"
         for V in "${VERSIONS[@]}"; do
           V=$(echo "$V" | xargs)  # trim whitespace
@@ -39,12 +41,16 @@ jobs:
           nvm use "$V" > /dev/null 2>&1
           VER=$(node --version)
 
+          # Dry-run install to collect EBADENGINE warnings without actually installing.
+          # awk extracts "package@version (node range)" from multi-line warning blocks.
+          # sort -u deduplicates (npm may warn about the same package multiple times).
           PKGS=$(npm ci --dry-run 2>&1 \
             | awk -F"'" '/EBADENGINE.*package:/{pkg=$2} /EBADENGINE.*node:/{if(pkg){print pkg" (node "$2")"; pkg=""}}' \
             | sort -u | tr '\n' ',' | sed 's/,$//' | sed 's/,/, /g')
 
           if [ -n "$PKGS" ]; then
             FAILED=true
+            # Escape pipe characters so they don't break the markdown table
             SAFE=$(echo "$PKGS" | sed 's/|/\\|/g')
             TABLE="$TABLE\n| $VER | ❌ Fail | $SAFE |"
           else
@@ -52,9 +58,11 @@ jobs:
           fi
         done
 
+        # Build the comment body
         if [ "$FAILED" = true ]; then ICON="❌"; else ICON="✅"; fi
         printf '%s\n## %s Engine Compatibility\n\n%b\n' "$MARKER" "$ICON" "$TABLE" > /tmp/comment.md
 
+        # Update existing comment (matched by HTML marker) or create a new one
         OLD=$(gh api "repos/${{ github.repository }}/issues/${PR}/comments" \
           --jq "[.[] | select(.body | contains(\"$MARKER\"))][0].id" 2>/dev/null || true)
         if [ -n "$OLD" ] && [ "$OLD" != "null" ]; then
@@ -63,6 +71,7 @@ jobs:
           gh pr comment "$PR" --body-file /tmp/comment.md
         fi
 
+        # Add "blocked" label on failure, remove it on success
         if [ "$FAILED" = true ]; then
           gh label create blocked --color D93F0B 2>/dev/null || true
           gh pr edit "$PR" --add-label blocked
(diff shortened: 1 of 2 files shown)


d3xter666:
**Note:** This flow can be ruesed accross repositories the following way:

```
uses: UI5/cli/.github/workflows/engine-compat.yml@main
with:
  node-versions: '20, 22, 24'
```

The caller uses `owner/repo/.github/workflows/file.yml@ref` syntax instead of `./.github/workflows/...`
No copying needed. The @ main can also be a tag (@ v1) or SHA for pinning. 


I gave it a try in an external repositorty and it seems to be working correctly: 
- https://github.com/d3xter666/test-ci/pull/1/changes#diff-86c810903f8631220438a8f9ec529ccb2ac3f52ca950cc95a999397077deb106R18
- https://github.com/d3xter666/test-ci/actions/runs/24188653031/job/70599746155

RandomByte:
This is a good idea, but I'm wondering how we can ignore dev-dependencies as discussed last week?

The [check-engine-light](https://github.com/textbook/check-engine-light/tree/main/packages/check-engine-light/src) implementation seems rather simple. If the lack of usage is really a concern for us we can also consider implementing such a tool ourselves. I had already started doing that, but then found that my name of choice "check-engine-light", was already taken by a tool with the very same functionality. In general, the check is really easy to implement and more efficient than the proposal here since we only need to parse the lockfile and not install any packages.

d3xter666:
> This is a good idea, but I'm wondering how we can ignore dev-dependencies as discussed last week?
> 
> The [check-engine-light](https://github.com/textbook/check-engine-light/tree/main/packages/check-engine-light/src) implementation seems rather simple. If the lack of usage is really a concern for us we can also consider implementing such a tool ourselves. I had already started doing that, but then found that my name of choice "check-engine-light", was already taken by a tool with the very same functionality. In general, the check is really easy to implement and more efficient than the proposal here since we only need to parse the lockfile and not install any packages.

If it's for the control of dependencies, we can leverage the `--omit` or `--include` flag.
Something that we also need to consider is how to redistribute this check accross all the repositories we own.
With this approach we easily update/modify the script in a single place and it automatically appears on the other repoisories. It can be "forzen" for certain repositories by providing a specific tag/hash/branch. Also this solution uses native and built-in tools that will always work. Direct access to PRs via `gh` CLI.

The issue with the listed packages are actually what we currently have- potential lack of update in case of security issues, not a major version (the major version release might not fit our engine).

From that perspective, if we want to use a custom script, we have the following options:
- Parse `package-lock.json` / `npm-shrinkwrap.json` (how to identify not prod dependencies?)
- Use shell with `npm ls --json [--omit <dev|optional|peer>]` command, so we can leverage Node's parsing directly without manually parsing anything or including dependencies like `arborist`, for example

Then we'll simply use `semver` in order to compare packages' versions against our `package.json`
Then, we have the following chalenges to address:
- What we want to do with this information?
    - Post to GitHub (maybe directly invoke `gh` in shell)
    - Exit code?
    - Something else
- How to redistribute the script?

d3xter666:
> > This is a good idea, but I'm wondering how we can ignore dev-dependencies as discussed last week?
> > The [check-engine-light](https://github.com/textbook/check-engine-light/tree/main/packages/check-engine-light/src) implementation seems rather simple. If the lack of usage is really a concern for us we can also consider implementing such a tool ourselves. I had already started doing that, but then found that my name of choice "check-engine-light", was already taken by a tool with the very same functionality. In general, the check is really easy to implement and more efficient than the proposal here since we only need to parse the lockfile and not install any packages.
> 
> If it's for the control of dependencies, we can leverage the `--omit` or `--include` flag. Something that we also need to consider is how to redistribute this check accross all the repositories we own. With this approach we easily update/modify the script in a single place and it automatically appears on the other repoisories. It can be "forzen" for certain repositories by providing a specific tag/hash/branch. Also this solution uses native and built-in tools that will always work. Direct access to PRs via `gh` CLI.
> 
> The issue with the listed packages are actually what we currently have- potential lack of update in case of security issues, not a major version (the major version release might not fit our engine).
> 
> From that perspective, if we want to use a custom script, we have the following options:
> 
> * Parse `package-lock.json` / `npm-shrinkwrap.json` (how to identify not prod dependencies?)
> * Use shell with `npm ls --json [--omit <dev|optional|peer>]` command, so we can leverage Node's parsing directly without manually parsing anything or including dependencies like `arborist`, for example
> 
> Then we'll simply use `semver` in order to compare packages' versions against our `package.json` Then, we have the following chalenges to address:
> 
> * What we want to do with this information?
>   
>   * Post to GitHub (maybe directly invoke `gh` in shell)
>   * Exit code?
>   * Something else
> * How to redistribute the script?

Just created a separate version with [check-engine-light](https://github.com/textbook/check-engine-light/tree/main/packages/check-engine-light/src): https://github.com/UI5/cli/pull/1360
Of course, we can quite simplify it and just reuse the `check-engine-light` package. I remember though that we discussed we just want the information whether certain package is ok or not. With the current setup we have the information available, but it's not blocking the PR from being merged.
If we use the pure `check-engine-light` package if everyhting is ok, then it will complete silently, but if there's an incompatibility it will exit with 1 and this will fail the general ci checks. This will have (almost) the same effect and hit the same boundaries as it is currently with `npm ci --engine-strict`

RandomByte:
>  I remember though that we discussed we just want the information whether certain package is ok or not.

I must have forgotten about that. What exactly is the expectation there? To get a list of packages that fail the engine check or more?

I thought we would simply add `check-engine-light` as an npm dependency and add an npm scripts that can be called from the CI. Why handle it different from the other tools like Knip?

d3xter666:
We have decided to go with the `check-engine-light`

PR #1360, opened 2026-04-14
ci: Add engine compatibility check for PRs (external package)

d3xter666:
JIRA: CPOUI5FOUNDATION-1225

`check-engine-light` is run as a regular CI step — if all dependencies are compatible it exits silently, and if any are incompatible it exits with code 1, failing the build.


.github/workflows/github-ci.yml
```diff
@@ -24,8 +26,12 @@ jobs:
       with:
         node-version: 22.20.0
 
+    # Check if "engines" field of dependencies is compatible with our Node.js version.
+    # This is a quick check to fail early if incompatible dependencies are introduced. 
+    - name: Check Node.js engine compatibility
+      run: npx -y check-engine-light
```
RandomByte: Could we add the package as a dependency to the root package.json together with a new npm script `check-engine`? https://github.com/UI5/cli/blob/main/package.json

d3xter666: Yep, sure! i forgot it! I'd need to also add it to the knip exclusion config

d3xter666: Done

RandomByte (dismissed):

RandomByte (approved):

codeworrior (approved):
Looks fairly simple now!

PR #1366, opened 2026-04-28
feat: Store metadata in SQLite

d3xter666:
JIRA: CPOUI5FOUNDATION-1209

Uses https://nodejs.org/api/sqlite.html to store incremental build metadata.
Gains write performance, since writing to SQL is non-blocking


d3xter666:
Already implemented in the main incremental build branch

PR #1374, opened 2026-05-08
fix(builder): Improve error ESM handling

RandomByte:
Up-port of https://github.com/SAP/ui5-builder/pull/1203

- **fix(builder): Reduce noise from ESM parse errors**
- **fix(builder): Skip ESM modules during bundling and log errors**
- **fix(builder): Also detect import.meta as ESM indicator**


matz3 (approved):
LGTM. I updated the PR title and commit messages in the PR body.

PR #1375, opened 2026-05-08
fix(builder): Update lbt/bundle/Resolver to ensure deterministic ordering of raw modules

RandomByte:
Upport of https://github.com/SAP/ui5-builder/pull/1208

When multiple raw modules share a common dependency, the topological
sort tie-breaks by input order. Previously, the input order depended
on async resolution timing (pool insertion order), making the output
non-deterministic.

Now modules are sorted by their filter definition order before
topological sorting, ensuring that modules at the same dependency
level appear in the order defined by the section's filter list.


codeworrior (approved):
Thank you for the up-port! LGTM.

PR #1385, opened 2026-05-11
ci: Run tests for workspace packages in parallel and as dedicated jobs

RandomByte:
- **ci: Add unit:summary script for parallel test runs with failure overview** - Useful for local development
- **ci: Run unit tests per workspace in parallel**
    - Useful in GitHub Actions to easily see which package has test failures. For instance, try finding the failing packages in output like this: https://github.com/UI5/cli/actions/runs/25683391554/job/75400615881?pr=1267

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,13 +12,20 @@ permissions:
   contents: read
 
 jobs:
-  test:
-    name: Unit and Integration
+  unit:
+    name: "${{ matrix.workspace }} (Node ${{ matrix.version }}, ${{ matrix.os }})"
     strategy:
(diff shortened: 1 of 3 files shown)


maxreichmann:
The CI changes definitely are going to be helpful 👍
I also really like the summary script. Would it be easy (or even useful) to add a --watch mode for automatic restarts?

d3xter666 (approved):

RandomByte:
> The CI changes definitely are going to be helpful 👍
> 
> I also really like the summary script. Would it be easy (or even useful) to add a --watch mode for automatic restarts?

I think that would be more complicated since we would need to parse the log messages instead of relying on the exit code. Feel free to try if you think it could be useful for you

Issue #512, opened 2021-03-30
Improve JSModuleAnalyzer to support local names for dependency detection


matz3:
## Is your feature request related to a problem? Please describe.

Currently, the UI5 Tooling is not able to detect all dependencies of a JavaScript module when it analyzes the code.
When using local names for certain APIs, such as `jQuery.sap.require` or `sap.ui.require`, the calls are not recognized and the dependencies are not detected.

## Describe the solution you'd like

The underlaying JSModuleAnalyzer (part of ui5-builder) should be improved in a way that allows recognition of the known APIs in case they are using a local variable name.

## Additional context

One example (there are many more use cases)

**Dependency to my/module.js is detected**
```js
sap.ui.define(["jquery.sap.global"], function() {
  jQuery.sap.require("my.module");
});
```

Dependency to my/module.js is **NOT** detected
```js
sap.ui.define(["jquery.sap.global"], function(jq) {
  jq.sap.require("my.module");
});
```

PR #1392, opened 2026-05-21
Feature: Depedency detection

EdrilanBerisha:
The `JSModuleAnalyzer` in `packages/builder/lib/lbt/analyzer/JSModuleAnalyzer.js` now detects module dependencies when known UI5 APIs are called through __local alias names__, not just their canonical names. Three patterns are supported:

1. __Factory function parameter aliasing `jquery.sap.global`:__

   ```js
   sap.ui.define(["jquery.sap.global"], function(jq) { jq.sap.require("my.module"); });
   ```

2. __Factory function parameter aliasing `sap/ui/core/Core`:__

   ```js
   sap.ui.define(["sap/ui/core/Core"], function(ui) { ui.require(["my/dep"]); ui.requireSync("other/dep"); });
   ```

3. __Top-level variable assignments:__

   ```js
   var jq = jQuery;  // or var jq = $; or var ui = sap.ui;
   jq.sap.require("my.module");
   ```

__Tests:__ 27 new unit tests were written in `JSModuleAnalyzer_aliasDetection.js` covering all alias patterns, conditional branches, position correctness guards, regression coverage for canonical calls, and fixture-file based tests. All 27 new tests + all 54 existing tests pass.

Feature was requested: [UI5/cli#512](https://github.com/UI5/cli/issues/512)


flovogt:
Thanks a lot for your contribution. We will discuss your enhancement request in the team and give an update here.

RandomByte:
Thank you for looking into this and contributing.

We had a closer look and have a few concerns before we can move forward. Issue #512 describes one specific pattern: a factory parameter aliasing `jquery.sap.global`. The PR goes considerably beyond that and adds three further patterns that weren't requested — and one of them is unfortunately incorrect: `sap/ui/core/Core` exports a `Core` instance, not the `sap.ui` namespace, so `core.require(...)` is not equivalent to `sap.ui.require(...)`. Treating it as such would produce false positives on existing UI5 code that calls methods on the Core instance.

The other concern is scope. `aliasMap` is a single map for the whole `analyze()` call, populated for every `VariableDeclarator` regardless of nesting and never cleared on scope exit. That means a `var jq = jQuery` in a nested closure leaks into the rest of the file, and any later `jq` parameter shadowing the alias gets misinterpreted. The analyzer already imports `escope`, but here a simpler approach should be sufficient: scope the alias map to the factory body of the `sap.ui.define` / `define` call that introduced it, so its lifetime matches the factory.

For a focused change that addresses the issue, I'd suggest:

- Limit the scope to the `jquery.sap.global` factory parameter case from the issue. If the var-assignment patterns turn out to be needed in practice, we can follow up separately.
- Inline the parameter→alias mapping into the existing `Syntax.CallExpression` branch where `sap.ui.define` / `define` is already handled, so it lives only for that factory body.
- Reuse `fromUI5LegacyName` / `fromRequireJSName` (already imported at the top of the file) instead of the new ad-hoc normalisation — those are the canonical helpers and will stay consistent with the rest of the analyzer.
- Drop the unused `JQUERY_SAP_GLOBAL_MODULE_IDS` constant and the unreachable `SAP_UI_CORE_MODULE_ID` comparison.
- Keep the tests that cover the jQuery factory-param case and add a shadowing negative test (e.g. an inner function with a parameter of the same name) to pin down the scope behaviour.

RandomByte:
Even after the review, this still looks like slop to me so I'm closing this PR for now.

I have doubts that you are personally interested in resolving https://github.com/UI5/cli/issues/512 (a five year old issue with no activity) and/or learning about the inner workings of this project. It rather feels like you are using a coding agent to try and make contributions to SAP OSS projects for your portfolio. We take contributions seriously and we generally review them with a human in the loop, so please be more considerate of our time.

Please reach out to me internally if I'm misjudging you and I will update this comment.

Regarding https://github.com/UI5/cli/issues/512: Please let us know if you are actually affected by this bug so we can prioritize a fix (and also understand the specific scope to resolve your case). In that case however, I think it would be more efficient for us to prompt a solution and iterate over it ourselves, rather than going through you. Unless you _are_ actually interested in learning about this project.

PR #1397, opened 2026-05-26
ci: Harden workflows

RandomByte:
- Scope release-please contents/pull-requests write permissions to the
  release-please job; publish jobs only need id-token: write
- Disable persist-credentials on checkouts that don't push to git
- Move repository.name interpolation into env var to avoid template
  injection in run blocks
- Replace spoofable github.actor check in dependabot-auto-merge with
  github.event.pull_request.user.login. Note: spoofing the dependabot
  actor alone is not sufficient to trigger the auto-merge step. The
  dependabot/fetch-metadata action only emits outputs for genuine
  dependabot PRs, so the merge step's check on
  steps.metadata.outputs.update-type would no-op on a spoofed run. The
  change closes the gap defensively.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -5,13 +5,14 @@ on:
     branches:
       - main
 
-permissions:
-  contents: write
-  pull-requests: write
+permissions: {}
 
 jobs:
   release-please:
     runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      pull-requests: write
     outputs:
       releases_created: ${{ steps.release.outputs.releases_created }}
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-please.yml
```diff
@@ -71,6 +72,8 @@ jobs:
         package: [logger, fs, builder, server, project]
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
```
matz3: Not needed as job does not have write permissions anyways, right?

RandomByte: Correct, updated 👍

matz3 (approved):

PR #1407, opened 2026-06-08
feat: Add liveReload

matz3:
Live reload automatically refreshes the browser whenever you change a source file in your project — no manual reload needed. This shortens the edit/test cycle during development.

JIRA: CPOUI5FOUNDATION-1224

### DIFF
--- a/internal/documentation/docs/pages/Server.md
+++ b/internal/documentation/docs/pages/Server.md
@@ -40,6 +40,7 @@ A project can also add custom middleware to the server by using the [Custom Serv
 | `csp` | See chapter [csp](#csp) |
 | `compression` | Standard [Express compression middleware](http://expressjs.com/en/resources/middleware/compression.html) |
 | `cors` | Standard [Express cors middleware](http://expressjs.com/en/resources/middleware/cors.html) |
+| `liveReloadClient` | See chapter [liveReload](#livereload) |
 | `discovery` |  See chapter [discovery](#discovery) |
 | `serveResources` | See chapter [serveResources](#serveresources) |
 | `testRunner` | See chapter [testRunner](#testrunner) |
@@ -66,6 +67,35 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
 ### discovery
 
 This middleware lists project files with URLs under several `/discovery` endpoints. This is exclusively used by the OpenUI5 test suite application.
+
+### liveReload
+
+Live reload automatically refreshes the browser whenever you change a source file in your project — no manual reload needed. This shortens the edit/test cycle during development.
+
+#### Usage
+
+Live reload is **enabled by default** with `ui5 serve`. Open your app in the browser, edit a source file, save — the page reloads.
+
+To control it:
+
+- **CLI flag**: `ui5 serve --live-reload` / `--no-live-reload`
+- **Project configuration** (specVersion 5.0+): `server.settings.liveReload` in `ui5.yaml`. See [Configuration](./Configuration.md).
+
+When the dev server is restarted, the browser automatically reconnects and reloads once the server is back. Saving multiple files at once triggers a single reload, not one per file.
+
+#### Technical Details
+
+The following describes how the middleware works internally. This is relevant for advanced users and custom middleware developers — not required for regular usage.
+
+When live reload is active, the UI5 server opens a WebSocket connection that notifies the browser of source changes and triggers a page reload.
+
+Reloads are driven by the `BuildServer`, which emits a debounced `sourcesChanged` event whenever watched source files change. A burst of changes therefore results in a single reload notification.
+
+The `liveReloadClient` middleware serves the client script at `/.ui5/liveReload/client.js`. The script tag is automatically injected into HTML responses by the `serveResources` middleware.
+
+To prevent intermediate proxies from idle-closing the WebSocket, the client sends a keepalive message every 30 seconds while the connection is open. The server echoes the same message back.
+
+When the WebSocket connection is lost (e.g. because the server was restarted), the client polls the WebSocket endpoint every second and reloads the page once the server accepts connections again. While the browser tab is hidden, polling pauses until it becomes visible again.
 
 ### serveResources
 This middleware resolves requests using the [ui5-fs](https://github.com/SAP/ui5-fs)-file system abstraction.
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -17,6 +17,8 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 - **Rename: Command Option `--cache-mode` is now `--snapshot-cache`**
 
+- **@ui5/server: Live Reload is enabled by default for `ui5 serve`**
+
 
 ## Node.js and npm Version Support
 
@@ -205,6 +207,19 @@ Delete the custom `test/Test.qunit.html` file from your test directory. This fil
 
 Depending on your project setup, you might need to update additional paths in configuration files or test runners to reflect the new structure.
 The test suite is now served under the standard `/test-resources/` path with the component's full namespace (e.g. `/test-resources/sap/ui/demo/todo/testsuite.qunit.html`).
+
+## Live Reload
+
+UI5 CLI v5 introduces a built-in [Live Reload](../pages/Server.md#livereload) feature for the development server. When running `ui5 serve`, the browser automatically reloads whenever project sources change. Live Reload is implemented via a WebSocket connection.
+
+Live Reload is **enabled by default**. It can be controlled via:
+
+- The new `--live-reload` CLI flag for `ui5 serve` (defaults to `true`). Pass `--no-live-reload` to disable it.
+- The new `server.settings.liveReload` configuration option in `ui5.yaml`. This setting is only available with [Specification Version 5.0](../pages/Configuration#specification-version-5-0) and higher.
+
+::: warning Custom Live Reload Middleware
+If your project uses a custom middleware that provides live reload functionality (e.g. [@sap-ux/reload-middleware](https://www.npmjs.com/package/@sap-ux/reload-middleware) or [ui5-middleware-livereload](https://www.npmjs.com/package/ui5-middleware-livereload)), the page may refresh more often than necessary when combined with the built-in feature. When upgrading, either remove the custom middleware or disable the built-in Live Reload via the `--no-live-reload` CLI flag or the `server.settings.liveReload` configuration option.
+:::
 
 ## Removal of Standard Server Middleware
 
--- a/packages/cli/lib/cli/commands/serve.js
+++ b/packages/cli/lib/cli/commands/serve.js
@@ -36,6 +36,13 @@ serve.builder = function(cli) {
 			default: false,
 			type: "boolean"
 		})
+		.option("live-reload", {
+			describe:
+				"Automatically reload the browser when project sources change. " +
+				"Overrides the 'liveReload' setting in the project's server configuration",
+			defaultDescription: "true",
+			type: "boolean"
+		})
 		.option("accept-remote-connections", {
 			describe: "Accept remote connections. By default the server only accepts connections from localhost",
 			default: false,
@@ -109,7 +116,7 @@ serve.builder = function(cli) {
 				"The 'Default' behavior is to invalidate the cache after 9 hours. 'Force' uses the cache only and " +
 				"does not create any requests. 'Off' invalidates any existing cache and updates from the repository",
 			type: "string",
-			defaultDescription: "Default", // Use "defaultdescription" to allow undefined (needed for evaluation)
+			defaultDescription: "Default", // Use "defaultDescription" to allow undefined (needed for evaluation)
 			choices: ["Default", "Force", "Off"],
 		})
 		.example("ui5 serve", "Start a web server for the current project")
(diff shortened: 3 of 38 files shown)


matz3:
To be decided: Should clients try to reconnect to the server, e.g. in case it has been restarted? This is currently not implemented, but I think it should be added.

packages/cli/lib/cli/commands/serve.js
```diff
@@ -165,11 +172,22 @@ serve.handler = async function(argv) {
 		}
 	}
 
+	let liveReload = argv.liveReload;
+	if (liveReload === undefined) {
+		const serverSettings = graph.getRoot().getServerSettings();
+		if (serverSettings && serverSettings.liveReload !== undefined) {
```
d3xter666, line 178: What about null, string (i.e. "false")?
Can we just check for true & false values

matz3:
> To be decided: Should clients try to reconnect to the server, e.g. in case it has been restarted? This is currently not implemented, but I think it should be added.

This has now been implemented.

packages/server/lib/liveReload/client.js
```diff
@@ -0,0 +1,113 @@
...
+		});
+	}
+
+	function waitForServer() {
+		return ping().then(function(ok) {
+			if (ok) {
+				return;
+			}
+			return wait(RECONNECT_INTERVAL_MS).then(loop);
+		});
```
d3xter666: This fragment is repeating. Can't we just wrap it in a function and reuse it

d3xter666 (commented):
LGTM

with small comments

packages/cli/lib/cli/commands/serve.js
matz3: The settings structure is guarded by the schema validation. For this reason, we do not additionally validate the values. Same already applies for the ports, which are taken as-is (see code above).

packages/server/lib/liveReload/client.js
matz3: True, good catch.

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,20 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
 
 This middleware lists project files with URLs under several `/discovery` endpoints. This is exclusively used by the OpenUI5 test suite application.
 
+### liveReload
+
+When live reload is active, `ui5 serve` opens a WebSocket connection that notifies the browser of source changes and triggers a page reload.
+
+Reloads are driven by the `BuildServer`, which emits a debounced `sourcesChanged` event (100ms debounce) whenever watched source files change. A burst of changes therefore results in a single reload notification.
```
RandomByte:
```suggestion
Reloads are driven by the `BuildServer`, which emits a debounced `sourcesChanged` event whenever watched source files change. A burst of changes therefore results in a single reload notification.
```

Maybe don't mention the exact number since that might change in the implementation and is not really relevant to the user.

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,20 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
...
+
+The `serveLiveReloadClient` middleware serves the client script at `/.ui5/liveReload/client.js`. The script tag is automatically injected into HTML responses by the `serveResources` middleware.
+
+To prevent intermediate proxies from idle-closing the WebSocket, the client sends a `{type: "ping"}` keepalive every 30 seconds while the connection is open. The server echoes the same message back.
```
RandomByte:
```suggestion
To prevent intermediate proxies from idle-closing the WebSocket, the client sends a keepalive message every 30 seconds while the connection is open. The server echoes the same message back.
```

Like above, the message content is an implementation detail that is not relevant in the user facing documentation

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -206,6 +208,15 @@ Delete the custom `test/Test.qunit.html` file from your test directory. This fil
 Depending on your project setup, you might need to update additional paths in configuration files or test runners to reflect the new structure.
 The test suite is now served under the standard `/test-resources/` path with the component's full namespace (e.g. `/test-resources/sap/ui/demo/todo/testsuite.qunit.html`).
 
+## Live Reload
+
+UI5 CLI v5 introduces a built-in [Live Reload](../pages/Server.md#livereload) feature for the development server. When running `ui5 serve`, the browser automatically reloads whenever project sources change. Live Reload is implemented via a WebSocket connection driven by the incremental build's `sourcesChanged` event.
```
RandomByte:
```suggestion
UI5 CLI v5 introduces a built-in [Live Reload](../pages/Server.md#livereload) feature for the development server. When running `ui5 serve`, the browser automatically reloads whenever project sources change. Live Reload is implemented via a WebSocket connection.
```

Event name is technical implementation detail that is not relevant in the user documentation

packages/server/lib/liveReload/client.js
```diff
@@ -0,0 +1,108 @@
...
+			if (msg.type === "reload") {
+				location.reload();
+			}
+			// Ignore other message types (e.g. "ping" echo)
+		} catch {
+			// ignore malformed messages
+		}
+	});
+
+	function ping() {
```
RandomByte:
```suggestion
	function tryReconnect() {
```

Misleading name

packages/server/lib/liveReload/client.js
```diff
@@ -0,0 +1,108 @@
...
+		return new Promise(function(resolve) {
+			let probe;
+			try {
+				probe = new WebSocket(wsUrl);
+			} catch {
+				resolve(false);
+				return;
+			}
+			function done(ok) {
+				probe.removeEventListener("open", onOpen);
```
RandomByte, line 60: Why do we have to remove these event listeners before closing the socket?

Without this the code would become simpler. We could use arrow-functions for the addEventListener calls

packages/server/lib/liveReload/server.js
```diff
@@ -0,0 +1,90 @@
...
+			// directions and don't idle-close the connection.
+			let type;
+			try {
+				type = JSON.parse(data.toString()).type;
+			} catch {
+				return;
+			}
+			if (type === "ping" && ws.readyState === ws.OPEN) {
+				ws.send(PING_MESSAGE);
+			}
```
RandomByte, line 93:
```suggestion
			try {
				const {type} = JSON.parse(data.toString());
				if (type === "ping" && ws.readyState === ws.OPEN) {
					ws.send(PING_MESSAGE);
				}
			} catch {
				return;
			}
```

packages/server/lib/middleware/MiddlewareManager.js
```diff
@@ -251,10 +252,25 @@ class MiddlewareManager {
 		});
 		await this.addMiddleware("compression");
 		await this.addMiddleware("cors");
+		await this.addMiddleware("serveLiveReloadClient", {
```
RandomByte: How about naming it simply `liveReload`?

RandomByte (changes requested):

packages/server/lib/liveReload/client.js
matz3: I intentionally did not use newer language features so that the client also works in older environments. Not sure whether this is actually relevant / needed. I thought of older UI5 LTS versions.

packages/server/lib/liveReload/server.js
matz3: This would hide exceptions raised by `ws.send`.

packages/server/lib/middleware/MiddlewareManager.js
matz3: The middleware only serves the `client.js`, so calling it `liveReload` might suggest that it also handles the liveReload WebSocket endpoint, which is not the case. Yes, the name is quite long, but it describes the purpose of the middleware. I'm open for better names, though.

packages/server/lib/liveReload/client.js
```diff
@@ -0,0 +1,108 @@
+(function() {
```
RandomByte, line 1: No unit tests for the client code 😿

packages/project/lib/build/BuildServer.js
```diff
@@ -308,10 +314,14 @@ class BuildServer extends EventEmitter {
 			this.#resourceChangeQueue.set(project.getName(), new Set([filePath]));
 		}
 
-		// : Emit event debounced
-		// Emit change event immediately so that consumers can react to it (like browser reloading)
-		// const changedResourcePaths = [...changes.values()].flat();
-		// this.emit("sourcesChanged", changedResourcePaths);
+		// Debounced emit so a burst of file changes results in a single reload notification
+		if (this.#sourcesChangedTimeout) {
```
RandomByte, line 318: Since we were talking about this behavior today: Do we have unit tests for this change and the debounce in particular?

RandomByte (dismissed):

packages/server/lib/liveReload/client.js
matz3: I didn't want to introduce real browser tests as I think it might be an overkill. But as it's based on WebSockets, my idea would be to mock the browser deps and run it in Node.js, as we can also connect via WebSocket from there. This would also make the coverage measurements much simpler.

packages/project/lib/build/BuildServer.js
matz3: Done

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,20 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
 
 This middleware lists project files with URLs under several `/discovery` endpoints. This is exclusively used by the OpenUI5 test suite application.
 
+### liveReload
+
+When live reload is active, the UI5 server opens a WebSocket connection that notifies the browser of source changes and triggers a page reload.
+
+Reloads are driven by the `BuildServer`, which emits a debounced `sourcesChanged` event whenever watched source files change. A burst of changes therefore results in a single reload notification.
```
flovogt, line 92: I really like the detailed documentation but I would recommend to have a more abstract / consumer friendly first section and then describing the technical details in a second section. The consumer has to know whats the benefit of this middleware and how to control it. IMHO The underlying technical aspects are only relevant for experts or custom middleware developers.

d3xter666 (commented):
LGTM

internal/documentation/docs/pages/Server.md
KlattG: Yes, that makes sense.

KlattG (changes requested):

internal/documentation/docs/pages/Server.md
matz3: Totally agree. There are now two sections.

KlattG (dismissed):

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,35 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
...
+#### Usage
+
+Live reload is **enabled by default** with `ui5 serve`. Open your app in the browser, edit a source file, save — the page reloads.
+
+To control it:
+
+- **CLI flag**: `ui5 serve --live-reload` / `--no-live-reload`
+- **Project configuration** (specVersion 5.0+): `server.settings.liveReload` in `ui5.yaml`. See [Configuration](./Configuration.md).
+
+When the dev server is restarted, the browser automatically reconnects and reloads once the server is back. Saving multiple files at once triggers a single reload, not one per file.
```
flovogt, line 84:
```suggestion
When the development server is restarted, the browser automatically reconnects and reloads the browser tab. Saving multiple files at once triggers a single reload, not one per file.
```

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,35 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
...
+
+#### Technical Details
+
+The following describes how the middleware works internally. This is relevant for advanced users and custom middleware developers — not required for regular usage.
+
+When live reload is active, the UI5 server opens a WebSocket connection that notifies the browser of source changes and triggers a page reload.
```
flovogt, line 90:
```suggestion
When live reload is active, the development server opens a WebSocket connection that notifies the browser of source changes and triggers a page reload.
```

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,35 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
...
+
+Reloads are driven by the `BuildServer`, which emits a debounced `sourcesChanged` event whenever watched source files change. A burst of changes therefore results in a single reload notification.
```
flovogt, line 92:
```suggestion
Reloads are driven by the `BuildServer` module, which emits a debounced `sourcesChanged` event whenever watched source files change. A burst of changes therefore results in a single reload notification.
```

internal/documentation/docs/pages/Server.md
```diff
@@ -67,6 +68,35 @@ With `serveCSPReports` set to `true`, the CSP reports are collected and can be d
...
+
+The `liveReloadClient` middleware serves the client script at `/.ui5/liveReload/client.js`. The script tag is automatically injected into HTML responses by the `serveResources` middleware.
+
+To prevent intermediate proxies from idle-closing the WebSocket, the client sends a keepalive message every 30 seconds while the connection is open. The server echoes the same message back.
+
+When the WebSocket connection is lost (e.g. because the server was restarted), the client polls the WebSocket endpoint every second and reloads the page once the server accepts connections again. While the browser tab is hidden, polling pauses until it becomes visible again.
```
flovogt, line 98:
```suggestion
When the WebSocket connection is lost (e.g. because the development server was restarted), the client polls the WebSocket endpoint every second and reloads the page once the server accepts connections again. While the browser tab is hidden, polling pauses until it becomes visible again.
```

flovogt (dismissed):

flovogt:
PM Review: LGTM

d3xter666 (approved):

maxreichmann (approved):

PR #1412, opened 2026-06-15
refactor(project): Add TaskDefinitions and task-level build signatures

RandomByte:
Extract the task-list construction out of TaskRunner into a new
TaskDefinitions class which owns the project-type to build-definition
switch, the standard task map, and the custom task validation and
de-duplication loop.

Each standard task in a build definition can expose an async
`determineBuildSignature()` returning a string contribution; the
fallback is `JSON.stringify(taskDef.options)`. Custom tasks expose the
same via the `determineBuildSignature` named export of their module,
surfaced through `Task#getDetermineBuildSignatureCallback`, and must
declare specVersion >= 5.0 to use it. `getProjectSignature` mixes the
aggregated string in alongside the existing inputs.

Split ProjectBuildContext into a synchronous wiring phase (logger,
TaskUtil, TaskDefinitions) and a static `create()` that resolves task
signatures, computes the project signature, and constructs the
ProjectBuildCache. `create()` parameters are reordered to
`(buildContext, project, baseSignature, cacheManager)`.

Add `TaskUtil#getReadOnlyInterface(specVersion)` for passing into the
`determineBuildSignature` callback before the task has run.

Also Include builder and fs version into project build signature.

### DIFF
--- a/packages/project/lib/build/TaskDefinitions.js
+++ b/packages/project/lib/build/TaskDefinitions.js
@@ -82,7 +82,7 @@ export default class TaskDefinitions {
 		return tasks;
 	}
 
-	async _addCustomTask(taskDef, tasks) {
+	_addCustomTask(taskDef, tasks) {
 		const project = this.#project;
 		const graph = this.#graph;
 
--- a/packages/project/lib/build/TaskDefinitions.js
+++ b/packages/project/lib/build/TaskDefinitions.js
@@ -1,0 +1,189 @@
+import crypto from "node:crypto";
+import {getLogger} from "@ui5/logger";
+
+export default class TaskDefinitions {
+	#graph;
+	#project;
+	#taskUtil;
+	#taskRepository;
+	#standardTasks;
+	#customTasks;
+	#initializingTasks;
+
+	constructor(graph, project, taskUtil, taskRepository) {
+		this.#graph = graph;
+		this.#project = project;
+		this.#taskUtil = taskUtil;
+		this.#taskRepository = taskRepository;
+	}
+
+	/**
+	 * Initializes the task list based on the project type
+	 *
+	 * This method:
+	 * 1. Loads the appropriate build definition for the project type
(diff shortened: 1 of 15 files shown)


packages/project/lib/build/helpers/ProjectBuildContext.js
```diff
@@ -34,37 +38,50 @@ class ProjectBuildContext {
...
 			cleanup: []
 		};
 	}
 
 	/**
-	 * Factory method to create and initialize a ProjectBuildContext instance
+	 * Factory method to create and fully initialize a ProjectBuildContext instance.
+	 *
+	 * Performs the async work that the constructor cannot: collecting task build signatures,
+	 * computing the project build signature, and constructing the {@link ProjectBuildCache}.
```
matz3: the link doesn't really work, does it?

packages/project/lib/build/TaskDefinitions.js
```diff
@@ -0,0 +1,189 @@
...
+
+	async _getCustomTasks() {
+		const projectCustomTasks = this.#project.getCustomTasks();
+		if (!projectCustomTasks || projectCustomTasks.length === 0) {
+			return new Map(); // No custom tasks defined
+		}
+		const tasks = new Map();
+		for (let i = 0; i < projectCustomTasks.length; i++) {
+			// Add tasks one-by-one to keep order as defined in project configuration
+			await this._addCustomTask(projectCustomTasks[i], tasks);
```
matz3, line 80: `_addCustomTask` does nothing async; could be changed to sync method, along with `_getCustomTasks`.

RandomByte: Done

packages/project/lib/build/helpers/ProjectBuildContext.js
RandomByte: Done

matz3 (approved):

PR #1415, opened 2026-06-16
refactor(project): Prevent concurrent writes to cache DB

matz3:
- Make APIs sync that do not need async behaviour
- Prevent concurrent access to DB (especially for batch transactions) by splitting async work from sync DB access

JIRA: CPOUI5FOUNDATION-1207

### DIFF
--- a/packages/project/lib/build/cache/BuildCacheStorage.js
+++ b/packages/project/lib/build/cache/BuildCacheStorage.js
@@ -21,8 +21,7 @@ export default class BuildCacheStorage {
 	#db;
 	#stmts;
 	#dbPath;
-	#inMetadataBatch = false;
-	#inContentBatch = false;
+	#inTransaction = false;
 
 	/**
 	 * @param {string} dbDir Directory in which to create the cache.db file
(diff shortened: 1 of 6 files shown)


packages/project/lib/build/cache/BuildCacheStorage.js
```diff
@@ -395,85 +394,47 @@ export default class BuildCacheStorage {
...
+				);
+			}
 			this.#db.exec("COMMIT");
+			this.#inTransaction = false;
+			return result;
+		} catch (err) {
+			try {
+				this.#db.exec("ROLLBACK");
+			} finally {
+				this.#inTransaction = false;
```
d3xter666, line 434: Maybe, this one needs to be in a `finally` clause, so the lock will always be cleaned up

matz3: If you mean `this.#inTransaction = false;` here, then it is already in a `finally` block, isn't it?

Or do you mean that we would only need one assignment, in a top-level `finally` block?
I think that would make sense and result in the same behavior.

d3xter666: My bad! Overlooked the netsed `try` block and thought it's `catch`

d3xter666 (approved):

PR #1418, opened 2026-06-17
ci: Update audit findings

matz3:


d3xter666 (approved):

codeworrior (approved):
New/modified entries look reasonable as they affect documentation (deps around vitepress) or tests (js-yaml) only - where the usage of the dependencies is under our control.

PR #1434, opened 2026-06-25
feat(cli)!: Scope project/workspace yargs options per command

matz3:
The yargs options --config / -c, --dependency-definition, --workspace-config,
and --workspace / -w were previously registered globally in base.js and shown
in --help for every command. Several commands silently ignored them:

  * ui5 config / init / versions did not load a project at all
  * ui5 add / remove / use only operate on ui5.yaml without workspace context

Move the option definitions into a new options.js helper module that exposes
applyProjectConfigOptions() and applyWorkspaceOptions(). Each command builder
now opts in to the option groups it actually consumes, and yargs' existing
strict mode rejects the options on commands that no longer accept them.

The shared last-value-wins dedupe coerce is exported as dedupeArray() so
commands can apply it to their own non-global options (framework-version,
dest, port, etc.) instead of relying on a central coerce list in base.js
that referenced options yargs only sees per-command.

BREAKING CHANGE: ui5 versions, ui5 init, and ui5 config no longer accept
--config, --dependency-definition, --workspace, or --workspace-config.
ui5 add, ui5 remove, and ui5 use no longer accept --workspace or
--workspace-config. Passing any of these to a command that does not
support them now fails with 'Unknown argument: <option>'.

### DIFF
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -15,7 +15,9 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 - **@ui5/cli: `ui5 init` defaults to Specification Version 5.0**
 
-- **Rename: Command Option `--cache-mode` is now `--snapshot-cache`**
+- **@ui5/cli: Option `--cache-mode` has been renamed to `--snapshot-cache`**
+
+- **@ui5/cli: Project/Workspace options are now scoped per command**
 
 - **@ui5/server: Live Reload is enabled by default for `ui5 serve`**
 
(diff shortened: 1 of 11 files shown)


RandomByte (approved):

RandomByte:
Regarding the commit message, it might be worth explicitly stating that the no longer accepted arguments already didn't have any effect in the past

matz3:
> Regarding the commit message, it might be worth explicitly stating that the no longer accepted arguments already didn't have any effect in the past

"Several commands silently ignored them" and the following bullet points explain that. Do you think that can be misunderstood?

RandomByte:
Sorry, I meant the BREAKING CHANGES paragraph. In the old setup, only that text would be visible right in the release note. Not sure how it is with release please now. But yeah, if you click through to the PR you'll find a good explanation 👍

PR #1436, opened 2026-06-25
docs: Add missing info about 'component' type

matz3:
Also removes the `generateVersionInfo` task from the `component` type, which was accidentally taken over from the application type.


flovogt:
@matz3 should we split the implementation from the documentation in 2 different commits? Actually its a fix, right?

RandomByte:
> @matz3 should we split the implementation from the documentation in 2 different commits? Actually its a fix, right?

This code did not exist in v4. Hence this is not a fix but rather a refactor imho. I'm fine with splitting the commits. Changes look good to me

RandomByte (approved):

matz3:
I think it's not worth splitting. I do not think this affects anyone.

PR #1439, opened 2026-06-29
feat: Add live status banner to ui5 serve

matz3:
JIRA: CPOUI5FOUNDATION-1275

**Light Mode:**
image

**Dark Mode:**
image

### DIFF
--- a/packages/cli/lib/cli/commands/serve.js
+++ b/packages/cli/lib/cli/commands/serve.js
@@ -1,10 +1,38 @@
 import path from "node:path";
 import os from "node:os";
-import chalk from "chalk";
 import baseMiddleware from "../middlewares/base.js";
 import {applyProjectConfigOptions, applyWorkspaceOptions, dedupeArray} from "../options.js";
+import {
+	REMOTE_CONNECTIONS_WARNING_LINES,
+	WARNING_STYLES,
+} from "../../serve/remoteConnectionsWarning.js";
 import {getLogger} from "@ui5/logger";
+import Logger from "@ui5/logger/Logger";
+import ConsoleWriter from "@ui5/logger/writers/Console";
+import {getVersion as getCliVersion} from "../version.js";
 const log = getLogger("cli:commands:serve");
+
+// Log levels that fall back to the plain-output path because the live banner
+// cannot represent the level's intent (firehose verbose logging) or because
+// the user has asked for silence entirely.
+const NON_BANNER_LEVELS = new Set(["perf", "verbose", "silly", "silent"]);
+
+// Collects all non-internal IPv4 addresses from the host's network
+// interfaces so the banner can list every reachable URL when the server
+// binds to all interfaces. Returns an empty array if no suitable address
+// is found.
+function findNetworkInterfaceAddresses() {
+	const interfaces = os.networkInterfaces();
+	const addresses = [];
+	for (const name of Object.keys(interfaces)) {
+		for (const iface of interfaces[name] ?? []) {
+			if (iface.family === "IPv4" && !iface.internal) {
+				addresses.push(iface.address);
+			}
+		}
+	}
+	return addresses;
+}
 
 // Serve
 const serve = {
@@ -135,6 +163,28 @@ serve.builder = function(cli) {
 };
 
 serve.handler = async function(argv) {
+	const useBanner =
+		process.stdout.isTTY === true &&
+		!NON_BANNER_LEVELS.has(Logger.getLevel());
+
+	let banner;
+	// Discover network addresses up front so the banner's initial paint can
+	// reserve the correct number of lines for the "Network:" section. Once the
+	// header is painted, swapping placeholder lines for real URLs without
+	// changing the line count keeps the live region from re-flowing.
+	const networkAddresses = (useBanner && argv.acceptRemoteConnections) ?
+		findNetworkInterfaceAddresses() : [];
+	if (useBanner) {
+		const {default: Banner} = await import("../../serve/Banner.js");
+		// Banner takes over all output
+		ConsoleWriter.stop();
+		banner = Banner.observe({
+			brand: {name: "UI5 CLI", version: getCliVersion() || ""},
+			acceptRemoteConnections: !!argv.acceptRemoteConnections,
+			networkAddressCount: networkAddresses.length,
+		});
+	}
+
 	const {graphFromStaticFile, graphFromPackageDependencies} = await import("@ui5/project/graph");
 	const {serve: serverServe} = await import("@ui5/server");
 	const {getSslCertificate} = await import("@ui5/server/internal/sslUtil");
@@ -154,6 +204,23 @@ serve.handler = async function(argv) {
 			snapshotCache: argv.snapshotCache ?? argv.cacheMode ?? "Default", // Use cacheMode as fallback
 			workspaceConfigPath: argv.workspaceConfig,
 			workspaceName: argv.workspace === false ? null : argv.workspace,
+		});
+	}
+
+	if (useBanner) {
+		// Fill in the project + framework section now that the graph has
+		// resolved — the rest of the header still shows placeholders until
+		// the server is bound.
+		const rootProject = graph.getRoot();
+		const frameworkName = rootProject.getFrameworkName?.();
+		const frameworkVersion = rootProject.getFrameworkVersion?.();
+		banner.setProject({
+			name: rootProject.getName(),
+			type: rootProject.getType(),
+			version: rootProject.getVersion(),
+			framework: frameworkName ?
+				{name: frameworkName, version: frameworkVersion} :
+				null,
 		});
 	}
 
@@ -215,27 +282,28 @@ serve.handler = async function(argv) {
 
 	const protocol = h2 ? "https" : "http";
 	let browserUrl = protocol + "://localhost:" + actualPort;
-	if (argv.acceptRemoteConnections) {
-		process.stderr.write("\n");
-		process.stderr.write(chalk.bold("⚠️  This server is accepting connections from all hosts on your network"));
-		process.stderr.write("\n");
-		process.stderr.write(chalk.dim.underline("Please Note:"));
-		process.stderr.write("\n");
-		process.stderr.write(chalk.bold.dim(
-			"* This server is intended for development purposes only. Do not use it in production."));
-		process.stderr.write("\n");
-		process.stderr.write(chalk.dim(
-			"* Vulnerable (custom-)middleware can pose a threat to your system when exposed to the network"));
-		process.stderr.write("\n");
-		process.stderr.write(chalk.dim(
-			"* The use of proxy-middleware with preconfigured credentials might enable unauthorized access " +
-			"to a target system for third parties on your network"));
-		process.stderr.write("\n\n");
+
+	if (useBanner) {
+		banner.setUrls({
+			local: browserUrl,
+			network: networkAddresses.length ?
+				networkAddresses.map((addr) => protocol + "://" + addr + ":" + actualPort) :
+				undefined,
+		});
+	} else {
+		if (argv.acceptRemoteConnections) {
+			process.stderr.write("\n");
+			for (const {text, style} of REMOTE_CONNECTIONS_WARNING_LINES) {
+				process.stderr.write((WARNING_STYLES[style] || WARNING_STYLES.bullet)(text));
+				process.stderr.write("\n");
+			}
+			process.stderr.write("\n");
+		}
+		process.stdout.write("Server started");
+		process.stdout.write("\n");
+		process.stdout.write("URL: " + browserUrl);
+		process.stdout.write("\n");
 	}
-	process.stdout.write("Server started");
-	process.stdout.write("\n");
-	process.stdout.write("URL: " + browserUrl);
-	process.stdout.write("\n");
 
 	if (argv.open !== undefined) {
 		if (typeof argv.open === "string") {
--- a/packages/cli/lib/cli/commands/serve.js
+++ b/packages/cli/lib/cli/commands/serve.js
@@ -2,10 +2,7 @@ import path from "node:path";
 import os from "node:os";
 import baseMiddleware from "../middlewares/base.js";
 import {applyProjectConfigOptions, applyWorkspaceOptions, dedupeArray} from "../options.js";
-import {
-	REMOTE_CONNECTIONS_WARNING_LINES,
-	WARNING_STYLES,
-} from "../../serve/remoteConnectionsWarning.js";
+import {REMOTE_CONNECTIONS_WARNING_LINES} from "../../serve/remoteConnectionsWarning.js";
 import {getLogger} from "@ui5/logger";
 import Logger from "@ui5/logger/Logger";
 import ConsoleWriter from "@ui5/logger/writers/Console";
@@ -293,8 +290,8 @@ serve.handler = async function(argv) {
 	} else {
 		if (argv.acceptRemoteConnections) {
 			process.stderr.write("\n");
-			for (const {text, style} of REMOTE_CONNECTIONS_WARNING_LINES) {
-				process.stderr.write((WARNING_STYLES[style] || WARNING_STYLES.bullet)(text));
+			for (const line of REMOTE_CONNECTIONS_WARNING_LINES) {
+				process.stderr.write(line);
 				process.stderr.write("\n");
 			}
 			process.stderr.write("\n");
--- a/packages/cli/lib/serve/Banner.js
+++ b/packages/cli/lib/serve/Banner.js
@@ -1,0 +1,365 @@
+import process from "node:process";
+import {createLogUpdate} from "log-update";
+import sliceAnsi from "slice-ansi";
+import chalk from "chalk";
+import Logger from "@ui5/logger/Logger";
+import {createInitialState, STATES} from "./state.js";
+import {renderHeader, renderStatusLine} from "./render.js";
+
+// Mirror the level prefixes used by @ui5/logger's default ConsoleWriter so
+// scrolled-back log lines look identical whether or not the banner is active.
+const LEVEL_PREFIX = {
+	warn: chalk.yellow("warn"),
+	error: chalk.bgRed.white("error"),
+};
+
+// Spinner tick interval while in `building` state.
+const BUILDING_TICK_MS = 120;
+
+/**
+ * Live banner for `ui5 serve`. Owns process.stdout while active.
+ *
+ * Subscribes to the @ui5/logger event surface (ui5.log, ui5.build-metadata,
+ * ui5.build-status, ui5.project-build-status, ui5.serve-status, ui5.log.stop-console)
+ * and renders a header + status line as a single live region. As more data
+ * becomes known (project resolved, server bound), the header is repainted
+ * in place — sections that aren't known yet show dim placeholders.
+ *
+ * The actual terminal bookkeeping (wrap-aware erase of the previous frame,
+ * cursor hide/show, screen-row vs. logical-line accounting) is delegated to
+ * the `log-update` package, which counts screen rows after wrap rather than
+ * logical lines — important when warnings exceed the terminal width.
+ *
+ * Lifecycle: {@link Banner.observe} paints the skeleton immediately (brand
+ * only, placeholders for the rest) and attaches the event listeners. Use it
+ * as soon as `ui5 serve` decides it wants a banner — well before the project
+ * graph is built — so the user sees feedback right away. Section setters
+ * ({@link Banner#setProject}, {@link Banner#setUrls}) then update individual
+ * header sections and repaint in place.
+ */
+class Banner {
+	#stdout;
+	#stopped = false;
+	#state;
+	#layout = {projectNameWidth: 0, taskNameWidth: 0};
+	#tickTimer = null;
+	#columns;
+	// Map of build-status: track which project index is next.
+	#projectOrder = [];
+	// `log-update` instance bound to `#stdout`. Owns the live region: every
+	// `#render()` call hands it the full multi-line frame and it diffs against
+	// the previous one, erasing wrap-correctly.
+	#logUpdate;
+
+	// Bound listeners so we can `process.off` them on stop().
+	#onLog;
+	#onBuildMetadata;
+	#onBuildStatus;
+	#onProjectBuildStatus;
+	#onServeStatus;
+	#onStopConsole;
+	#onResize;
+
+	/**
+	 * Paint the banner skeleton immediately (with placeholders for any
+	 * unknown sections) and attach the event listeners. Subsequent section
+	 * setters refine the skeleton in place.
+	 *
+	 * @param {object} [opts]
+	 * @param {WriteStream} [opts.stdout]
+	 * @param {object} [opts.brand]        { name, version } — usually known up front
+	 * @param {boolean} [opts.acceptRemoteConnections] known up front from argv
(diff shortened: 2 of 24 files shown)


d3xter666 (commented):
LGTM

Tried different scenarios with file modifications, deleteions, new files and it seems to be working as expected

packages/logger/lib/loggers/Serve.js
```diff
@@ -0,0 +1,99 @@
...
+		}
+		const level = "info";
+		const hasListeners = this._emit(Serve.SERVE_STATUS_EVENT_NAME, {
+			level,
+			status: "serve-build-done",
+			hrtime,
+		});
+		if (!hasListeners) {
+			// Plain-text fallback for the no-listener path. The banner formats
+			// via pretty-hrtime; this stays inline to avoid pulling a new
```
RandomByte, line 71: Since we use `pretty-hrtime` in other packages, IMHO it would be fine to also use it here - or are there greater concerns?

packages/cli/lib/serve/Banner.js
```diff
@@ -0,0 +1,514 @@
+import process from "node:process";
```
RandomByte, line 1: I'm wondering whether this should be located in the cli package or rather the logger. The `Console` writer is located in the logger package, so I would have expected this to be located there as well. What's the reason to have it in the CLI?

packages/cli/lib/serve/render.js
```diff
@@ -0,0 +1,173 @@
...
+	const lines = [];
+	lines.push("");
+	const brandVersion = brandInfo?.version ? chalk.dim("v" + brandInfo.version) : "";
+	lines.push(`  ${brand(brandInfo?.name || "UI5 CLI")} ${brandVersion}`);
+	lines.push("");
+
+	const localStr = urls?.local ?
+		accent(urls.local) :
+		placeholder("binding…");
+	lines.push(`  ${arrow} ${accentBold("Local:")}    ${localStr}`);
```
RandomByte: If you agree, maybe we can drop the two-space indent. I think we should make full use of the available space and not add padding where it is not really necessary for readability

packages/cli/lib/serve/render.js
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	// `acceptRemoteConnections` reflects user intent (the flag was passed), so
+	// the warning is meaningful from the first paint — well before the server
+	// has actually bound to the network.
+	if (acceptRemoteConnections) {
+		lines.push("");
+		for (const {text, style} of REMOTE_CONNECTIONS_WARNING_LINES) {
+			lines.push("       " + (WARNING_STYLES[style] || WARNING_STYLES.bullet)(text));
```
RandomByte: Is it necessary to add such a great indent? Looks a bit odd to be and not really aligned with the general layout

packages/cli/lib/cli/commands/serve.js
```diff
@@ -1,11 +1,39 @@
...
+// Log levels that fall back to the plain-output path because the live banner
+// cannot represent the level's intent (firehose verbose logging) or because
+// the user has asked for silence entirely.
+const NON_BANNER_LEVELS = new Set(["perf", "verbose", "silly", "silent"]);
+
+// Collects all non-internal IPv4 addresses from the host's network
+// interfaces so the banner can list every reachable URL when the server
+// binds to all interfaces. Returns an empty array if no suitable address
+// is found.
+function findNetworkInterfaceAddresses() {
```
RandomByte, line 21: Could we move this to the Banner implementation? It's not clear to me why this needs to happen in the serve command handler. `Banner.observe` receives the _count_ of addresses and only later we provide a list of "network". I feel like these are Banner specific implementation details and all it needs to know from the command handler is whether remote connections are accepted and later the server port (which _could_ also be submitted via `loggers/Serve` and a dedicated event I guess?)

packages/cli/lib/cli/commands/serve.js
```diff
@@ -135,6 +163,28 @@ serve.builder = function(cli) {
...
+	// Discover network addresses up front so the banner's initial paint can
+	// reserve the correct number of lines for the "Network:" section. Once the
+	// header is painted, swapping placeholder lines for real URLs without
+	// changing the line count keeps the live region from re-flowing.
+	const networkAddresses = (useBanner && argv.acceptRemoteConnections) ?
+		findNetworkInterfaceAddresses() : [];
+	if (useBanner) {
+		const {default: Banner} = await import("../../serve/Banner.js");
+		// Banner takes over all output
+		ConsoleWriter.stop();
```
RandomByte, line 177: We only initialized it in middlewares/logger.js, a bit strange to then stop it here again (basically we could have just never started it). Could we initialize the banner in the logger middleware instead?

packages/cli/lib/serve/render.js
```diff
@@ -0,0 +1,173 @@
...
+// Render the status line as a single string (no trailing newline).
+export function renderStatusLine(state, layout = {}) {
+	const label = `  ${chalk.dim("Status")}    `;
+	switch (state.state) {
+	case STATES.INITIAL:
+		return `${label}${chalk.dim("○")} ${chalk.dim(pad("starting"))}`;
+	case STATES.READY: {
+		let suffix = "";
+		if (state.lastBuildHrtime) {
+			suffix = ` ${chalk.dim("·")} ${chalk.dim("built in " + prettyHrtime(state.lastBuildHrtime))}`;
```
RandomByte:
```suggestion
			suffix = ` ${chalk.dim("·")} ${chalk.dim("ready in " + prettyHrtime(state.lastBuildHrtime))}`;
```

It's possible that no build occured

packages/cli/lib/serve/Banner.js
```diff
@@ -0,0 +1,514 @@
...
+		process.off("ui5.project-build-status", this.#onProjectBuildStatus);
+		process.off("ui5.serve-status", this.#onServeStatus);
+		process.off("ui5.log.stop-console", this.#onStopConsole);
+		if (typeof this.#stdout.off === "function") {
+			this.#stdout.off("resize", this.#onResize);
+		}
+	}
+
+	#handleLog({level, message, moduleName}) {
+		// The banner curates `info` away (the status line already represents
```
RandomByte, line 294: Why `info` "curated away" here? It might contain important information (e.g. when installing framework dependencies) that is now only visible with log level verbose.

Maybe there was a misunderstanding here? The build state is not logged automatically if event handlers are registered, so this shouldn't be a concern AFAIK.

packages/cli/lib/serve/Banner.js
```diff
@@ -0,0 +1,514 @@
...
+import {createLogUpdate} from "log-update";
+import sliceAnsi from "slice-ansi";
+import chalk from "chalk";
+import Logger from "@ui5/logger/Logger";
+import {createInitialState, STATES} from "./state.js";
+import {renderHeader, renderStatusLine} from "./render.js";
+
+// Mirror the level prefixes used by @ui5/logger's default ConsoleWriter so
+// scrolled-back log lines look identical whether or not the banner is active.
+const LEVEL_PREFIX = {
```
RandomByte, line 16: Could reuse `writers/Console #getLevelPrefix`

packages/cli/lib/serve/Banner.js
```diff
@@ -0,0 +1,514 @@
...
+			`${levelPrefix} ${message}`;
+		this.logAbove(formatted);
+	}
+
+	#handleBuildMetadata({projectsToBuild}) {
+		this.#projectOrder = Array.from(projectsToBuild);
+		this.#state.totalProjects = this.#projectOrder.length;
+		this.#state.currentProjectIndex = 0;
+		this.#state.currentProjectName = "";
+		this.#state.currentTaskName = "";
```
RandomByte, line 318: It seems like this initialization work is already done in the `serve-building` state handler of the `ui5.serve-status` event. Maybe we don't need to do it again here? It's just a bit hard to understand why it's done twice if that's really necessary.

RandomByte (changes requested):
Some of these we could also tackle in a later follow-up. E.g. moving the implementation into the logger package (if we actually want to do that) is not a top priority for me and might come with extra work since we then need to account for multiple versions of the logger package being in use (which however is a good reason to couple the banner to the Serve-logger implementation)

packages/logger/lib/loggers/Serve.js
matz3: As it's rather the fallback use-case I thought it is not worth adding the dependency also to the logger. But yeah, we already use it, so it would not be overhead.

packages/cli/lib/serve/Banner.js
matz3: My reasons for CLI were:
- It has knowledge about a lot of project, builder, server functionality/data, so it would be the right level as CLI depends on them.
- CLI is currently the only place where we need this, unless we would also want to make it available for API usage, which I think should not be a priority right now.
- "accepting remote connections" warning can easily be shared between the old and the new output, which otherwise would be more complex, and should IMHO not be something known to the logger.

packages/cli/lib/cli/commands/serve.js
matz3: The idea is to separate the data layer from the presentation layer. Banner does not care how the interface addresses are retrieved. But of course, we do not have many different ways, so right now it does not matter where the code lives.
Yes, using events for the data instead of the setters would also be an option. I think it evolved that way, as I initially did not have the placeholders, and only rendered once everything was ready.

packages/cli/lib/serve/Banner.js
matz3: Yes, progress of framework library installation is currently not visible in the banner. I already have a todo note for that, but thought that it would be okay without it in an initial version.

packages/cli/lib/serve/render.js
matz3: Then it would be `● ready    · ready in 1.23 s`. I would then rather drop the `built` or even `built in` completely:

```
Status    ● ready    · in 1.23 s
```
or
```
Status    ● ready    · 1.23 s
```

Or maybe `took` ?
```
Status    ● ready    · took 1.23 s
```

flovogt: @KlattG whats your opinion here?

packages/cli/lib/serve/Banner.js
matz3: We decided to keep the code in cli for now, but with a TODO comment to move it into logger in future. Having it in the cli allows us to move it again without having to deal with incompatible changes.

packages/cli/lib/serve/render.js
KlattG: To keep the message ambiguous in case no build occurred, how about the following:

```
Status    ● ready    · Duration: 1.23 s
```

packages/cli/lib/cli/commands/serve.js
matz3: It is not ideal, but as the current approach does not seem to cause any issues by starting and stopping the ConsoleWriter, we agreed to keep it like that for now.

packages/cli/lib/cli/commands/serve.js
matz3: We agreed to keep it like it is for now, but I added a TODO comment to Banner.js to state that things should be improved in future.

packages/cli/lib/serve/render.js
KlattG: Or maybe even better, how about this?

```
Status    ● ready    · Time elapsed: 1.23 s
```

matz3: We agreed on `took` for now, but I will create a follow-up PR to address some findings, so I will apply any further suggestions in there.

packages/cli/lib/serve/Banner.js
matz3: Will be improved in a follow-up PR.

packages/cli/lib/serve/Banner.js
matz3: Will be improved in a follow-up PR.

packages/cli/lib/serve/Banner.js
matz3: Yes, that's right. The code is currently more defensive to recover from unexpected state changes in which some fields could still be populated. I do not think that it causes much overhead or complicate the code too much, so I would just leave it as it is in this PR, and consider for the follow-up work.

A better improvement would be to have some functions to manage the state, which would then combine for example those 3 lines, which are currently duplicated code.

matz3:
I applied some changes, especially improvements in the rendering (more compact), and using the `figures` packages for unicode symbols with fallback.

More changes will be applied in a follow-up PR, as we already have two other improvements building on top of this PR, which would then be unblocked by merging this now.

packages/cli/lib/serve/render.js
matz3: @KlattG - sorry, I didn't see your comments when submitting mine. Thank you! I now applied `Time elapsed: 1.23 s`.

We would merge this PR soon, to unblock other related changes.

Would you mind also having a look at the `docs` changes in this PR?
The banner console output can best be checked and reviewed by using UI5 CLI and starting a server, but I could also provide screenshots if needed.
Alternatively, most texts are part of the `render.js` file: https://github.com/UI5/cli/pull/1439/changes#diff-bb6d62110576db3927f93cffdd5d9d264c5698050bef21b35d82f44142aced7e

marianfoo:
Just out of curiosity, what would happen if I use cap-plugin-ui5 and serve multiple UI5 Apps at the same time using CAP?
What I see in this PR I can only set output to silent to avoid this?

matz3:
> Just out of curiosity, what would happen if I use cap-plugin-ui5 and serve multiple UI5 Apps at the same time using CAP? What I see in this PR I can only set output to silent to avoid this?

I would assume nothing would change. `cap-plugin-ui5` does not seem to invoke the CLI via `ui5 serve`, so this feature won't be active.

packages/cli/lib/serve/state.js
```diff
@@ -0,0 +1,51 @@
+// Pure state holder for the live banner.
+//
+// All mutations go through the `update` function so callers can subscribe to
```
codeworrior: Is this comment outdated? I see no `update` function. Changes seems to be handled individually and then `Banner.prototype.#render` is called.

packages/cli/lib/serve/state.js
```diff
@@ -0,0 +1,51 @@
...
+// state changes and trigger redraws without sprinkling redraw calls across
```
codeworrior: well, this "sprinkling" rather seems to be the truth?

packages/cli/lib/serve/state.js
matz3: Yes, I missed that. Thanks!

codeworrior:
I've tested it locally on a bigger project where I use the workspace feature of `ui5 serve`. The new banner reports the framework version of the project
```sh
UI5 CLI v5.0.0-alpha.5

❯ Local:   http://localhost:8080
❯ Network: use --accept-remote-connections to expose

Project   sapui5-dist-testsuite  (application)  v1.150.0-SNAPSHOT
Framework SAPUI5 1.150.0-SNAPSHOT

Status    ● ready    · Time elapsed: 14 s
```
but I get no indication that this doesn't match the workspace content. With the old log output, I saw the version of each project contained in the workspace sources:
```sh
info graph:helpers:ui5Framework Using SAPUI5 version: 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.core via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.layout via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.unified via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.codeeditor via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.dt via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.suite via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.m via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.table via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.commons via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ndc via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.fl via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.mdc via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.tnt via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.ux3 via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.f via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.rta via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.support via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.integration via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.uxap via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.testrecorder via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project themelib_sap_belize via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project themelib_sap_fiori_3 via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project themelib_sap_horizon via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project themelib_sap_bluecrystal via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.webanalytics.core via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.makit via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.me via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.export via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.richtexteditor via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.viz via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.suite.ui.microchart via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.vbm via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ca.ui via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.rules.ui via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.vtm via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.zen.commons via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.zen.crosstab via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.uiext.inbox via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.sac.df via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.documentation via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.chart via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.insights via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ca.scfld.md via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.comp via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.suite.ui.commons via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.vk via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.gantt via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.generic.template via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.collaboration via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ui.generic.app via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ovp via local workspace to version 1.149.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.zen.dsh via local workspace to version 1.150.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.ushell via local workspace to version 1.151.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.apf via local workspace to version 1.148.0-SNAPSHOT
info graph:helpers:ui5Framework Resolved project sap.suite.ui.generic.template via local workspace to version 1.151.0-SNAPSHOT

```

As I appreciate the focused information presented by the banner, I would not like to always switch it off, just to see those inconsistencies. But I also don't want to hunt ghost bugs that in reality are unnoticed source inconsistencies.

Would it be possible to show an indicator for the inconsistencies, e.g. after the framework version, or is it too costly to determine this?
```sh
UI5 CLI v5.0.0-alpha.5

❯ Local:   http://localhost:8080
❯ Network: use --accept-remote-connections to expose

Project   sapui5-dist-testsuite  (application)  v1.150.0-SNAPSHOT
Framework SAPUI5 1.150.0-SNAPSHOT (not consistent with workspace content)

Status    ● ready    · Time elapsed: 14 s
```

matz3:
@codeworrior thank you for testing it and bringing this up! The downloading status of the framework packages is something I still wanted to add, but I didn't think about the workspaces. Would you prefer to only have such an inconsistency check, or would you like to be able to get back the full list of packages that are resolved via the workspace?

I feel like we might need an option to have a more verbose output in the banner, but far less compared to what `--verbose` prints. In the banner, we could try to include the most important information, such as the inconsistency warning, how many framework libs were resolved from the workspace (e.g. `9 of 10 projects resolved via workspace`), etc., but without always listing everything.

As stated earlier, I would still like to merge this PR now, to unblock other work. I will address this in a follow-up PR.

codeworrior:
I also see this as a potential follow-up, not as mandatory for an initial feature commit.

Reg. my expectations, I'm a bit undecided. "Dreaming", I would like to a see a summary, maybe with numbers as you suggested, maybe with an inconsistency indicator). And with an option or key press, see the full list of resolved projects / versions.

RandomByte (approved):
As agreed, let's merge this as is and continue working on the mentioned points in follow-up PRs

PR #1442, opened 2026-07-01
ci: Add aggregator job for unit test matrix

RandomByte:
Add a 'Tests' job that depends on the full unit test matrix so branch
protection can require a single check instead of every workspace/Node/OS
combination. The 36 matrix rows remain visible in the PR checks list;
only the required-check surface collapses.

'if: always()' ensures the aggregator still runs when matrix cells fail,
so a failure surfaces as 'failed' rather than 'skipped'.


matz3 (approved):
That sounds like a great solution 👍🏻

PR #1445, opened 2026-07-03
refactor: Move ui5 serve banner from cli to logger

matz3:
- Relocate banner rendering and state tracking out of `@ui5/cli` into a new `InteractiveConsole` writer in `@ui5/logger`
- Drive the banner via `ui5.tool-mode` / `ui5.tool-info` log events instead of direct CLI calls
- Split writer internals into `render`, `format`, and per-region `state/*` modules
- Emit project/server lifecycle events from `@ui5/project` and `@ui5/server` to feed the writer
- Reorder regions so project metadata renders before URLs — better fits future use cases like ui5 build, where the project is resolved before work begins

This PR covers several review comments from #1439, but not all of them.
For example, better handling of the log level and framework resolution/download progress will be tackled in another PR.


matz3:
I pushed a tmp branch to allow including the initial PR #1439 into the diff: https://github.com/UI5/cli/compare/main-before-serve-banner...ui5-serve-console-output-followup

This should make a review easier as it allows to see what effectively changed, not only what the follow-up PR does.

packages/logger/lib/writers/interactiveConsole/render.js
```diff
@@ -0,0 +1,182 @@
...
+		const projectType = project.type ? chalk.dim(`(${project.type})`) : "";
+		const projectVersion = project.version ? chalk.dim("v" + project.version) : "";
+		lines.push(`${chalk.dim("Project")}   ${chalk.bold(project.name)}` +
+			(projectType ? `  ${projectType}` : "") +
+			(projectVersion ? `  ${projectVersion}` : ""));
+		if (framework && framework.name) {
+			const frameworkVersion = framework.version ? ` ${framework.version}` : "";
+			lines.push(`${chalk.dim("Framework")} ${chalk.bold(framework.name + frameworkVersion)}`);
+		} else {
+			lines.push(`${chalk.dim("Framework")} ${placeholder("(none)")}`);
```
RandomByte: I was wondering whether we should just omit the line if no framework version is configured. Obviously for a UI5 app to work, there is a framework. So it's not really "none" but rather "unknown" - in which case we might just omit the info (and render an empty line) to avoid confusion?

RandomByte (dismissed):
Really much better, thanks! Just one optional comment

packages/logger/lib/writers/interactiveConsole/render.js
codeworrior: Due to the use of the fiori-tools-proxy middleware, this might be a very common scenario (_Claude found 4169 out of 6505 ui5*.yaml files in my local cluecode repository which do not contain a framework config_). 
Could a middleware like fiori-tools-proxy in future contribute this information, e.g. via some log point?

matz3: Done. It now renders no line for placeholder, and adds an additional line only if there is a framework configured. I think this is better, as we might have more optional info in future, so rendering empty lines for them would look quite odd. If the "flickering" is too disturbing, we should rather not render any placeholders below, until we have resolved the project details.
My main goal of the placeholders was to render _something_ initially, and not only after the server is listening.

RandomByte (approved):

PR #1450, opened 2026-07-06
refactor: Add cache validation methods for use in ui5 serve

RandomByte:
Main change: `refactor(project): Validate dependency caches in background after each build`

After server startup (and initial project build), and while no HTTP requests have been received yet, this will initialize and validate the ProjectBuildCache instances of all projects in the dependency tree. The user will see this activity as a new `VALIDATING` server state in the interactive console. The resulting state will be either "stale" or "ready", depending on whether the restored cache is "valid" (i.e. can be used without a build) or not.

The other commits are cleanups on top of this: consolidation of BuildServer state transitions, collapse of the two cache-validation entry points on `ProjectBuildContext`, deduplication of `ProjectBuilder.#build` / `#validate`, a guard against a concurrent build claim in `#triggerRequestQueue`, plus a matching pair of logger extractions and test-scaffolding helpers.

### DIFF
--- a/packages/project/lib/build/ProjectBuilder.js
+++ b/packages/project/lib/build/ProjectBuilder.js
@@ -340,6 +340,46 @@ class ProjectBuilder {
 	}
 
 	/**
+	 * Resolves the project build contexts for the requested projects (plus their transitive
+	 * build-time dependencies) and returns them in dependency-first (DFS) order. Shared setup
+	 * for {@link #build} and {@link #validate}; the DFS ordering ensures that dependencies are
+	 * processed before their dependents, so propagated resource changes reach dependents in
+	 * time.
+	 *
+	 * @param {string[]} requestedProjects Names of the projects the caller wants to process
+	 * @returns {Promise<{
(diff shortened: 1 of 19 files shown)


packages/project/lib/build/ProjectBuilder.js
```diff
@@ -425,6 +477,97 @@ class ProjectBuilder {
...
+						usesCache = await projectBuildContext.validateCache();
+						if (this.#log.isLevelEnabled("perf")) {
+							this.#log.perf(
+								`validateCache for ${projectName} ` +
+								`completed in ${(performance.now() - valStart).toFixed(2)} ms ` +
+								`(usesCache=${usesCache})`);
+						}
+					}
+
+					if (projectValidatedCallback && isRequested) {
```
matz3, line 557: `signal?.throwIfAborted();` should be done before calling the callback. This ensures consistency with the similar code in `#build`.

matz3 (approved):

packages/project/lib/build/ProjectBuilder.js
RandomByte: Done

PR #1453, opened 2026-07-16
refactor(builder): Create `sap-ui-version.json` by default

maxreichmann:
JIRA: CPOUI5FOUNDATION-1157

---

**What has been done?**
- Exclusion of task `generateVersionInfo`removed
   - now it will be executed by default for every **build** of project type **`application`** (default, jsdoc, self-contained)
- Tests refactored accordingly
   - +new test verifying exclusion/inclusion of `generateVersionInfo`: [ProjectBuilder.integration.js:2758](https://github.com/UI5/cli/blob/cd38857a4f64a1353f27db6893643d23535fbad8/packages/project/test/lib/build/ProjectBuilder.integration.js#L2758)
- Docs
   - `migrate-v5.md`:  new section explaining new behavior
   - `Builder.md`: changed `generateVersionInfo` from disabled to enabled in Standard Tasks table

**Info**
- for **Server**:
   - task `generateVersionInfo` is explicitly NOT executed because middleware `versionInfo` IS executed (which creates `sap-ui-version.json` anyways)
   - Test verifying exclusion of `generateVersionInfo`: [BuildServer.integration.js:1030](https://github.com/UI5/cli/blob/f354410161f654f4622a9ef14c5c3f91f8b33de6/packages/project/test/lib/build/BuildServer.integration.js#L1030)
   - Tests verifying `generateVersionInfo` is added to the `excludedTasks` array: [server.js:143](https://github.com/UI5/cli/pull/1453/changes#diff-5ee6bf8d5a61e10deb932168617dbec4e094ab2c529812607a6b46b010917ea8R143-R226)
   - (existing) Test verifying `sap-ui-version.json` is generated: [server/test/lib/server/main.js:204](https://github.com/UI5/cli/blob/f354410161f654f4622a9ef14c5c3f91f8b33de6/packages/server/test/lib/server/main.js#L204)


matz3:
We should also check how this now relates to the `versionInfo` middleware. My assumption would be that we need to keep the middleware to serve the file for non-application projects (and when the task was skipped?). For application projects, we should be able to just serve the file from the application build output. This should at least be tested, confirmed and documented.

maxreichmann:
Update:
- `generateVersionInfo` is now excluded from server builds (for normal builds, it's executed by default).
- Tests are added and existing ones are adjusted.
- Docs are adjusted.
--> Main part is done and ready for review.

---

"Side tasks done": I'm also fine to move those changes out to another PR if this eases the review process.

internal/documentation/docs/pages/Builder.md
```diff
@@ -23,31 +23,31 @@ A project can add custom tasks to the build by using the [Custom Tasks Extensibi
...
+| transformBootstrapHtml                                                                                 | *disabled* <sup>3</sup> |                         |                         |                         |
+| [generateBundle](../api/module-@ui5_builder_tasks_bundlers_generateBundle)                             | *disabled* <sup>4</sup> | *disabled* <sup>4</sup> | *disabled* <sup>4</sup> |                         |
+| [buildThemes](../api/module-@ui5_builder_tasks_buildThemes)                                            |                         |                         |         enabled         |         enabled         |
+| [generateThemeDesignerResources](../api/module-@ui5_builder_tasks_generateThemeDesignerResources)      |                         |                         | *disabled* <sup>5</sup> | *disabled* <sup>5</sup> |
+| [generateVersionInfo](../api/module-@ui5_builder_tasks_generateVersionInfo)                            |         enabled         |       *disabled*        |       *disabled*        |       *disabled*        |
```
matz3: If I see the diff correctly, the `*disabled*` for the types other than application have been added. Previously, the cells were empty, which means the task can never be executed for that project type (based on the build definition, like `packages/project/lib/build/definitions/application.js`), even if explicitly enabled by the caller.
I do not see code changes that change this situation, and I think the previous state still holds true: During a build, we only support running the `generateVersionInfo` for application projects, and it can not be enabled for other project types.

matz3: Should we add a foot-note for the server exception also here? This page is in context of the builder, but as we also run a build on the dev server, I think it would be worth noting here as well.

packages/project/lib/graph/ProjectGraph.js
```diff
@@ -768,6 +768,15 @@ class ProjectGraph {
 		cache = Cache.Default,
 		ui5DataDir,
 	}) {
+		// Explicitly exclude task "generateVersionInfo" for Server builds
+		// because middleware "versionInfo" will generate the version info anyways.
```
matz3: This handling should be moved to into the `@ui5/server` code in `packages/server/lib/server.js`, which does provide an HTTP server with the mentioned `versionInfo` middleware.

This `@ui5/project` API only returns the BuildServer, which is the underlying functionality, but it is not opinionated about the integration, and does not know about the `versionInfo` middleware.

This means that the BuildServer integration tests are likely unaffected by this change, which is fine.

matz3 (commented):
> "Side tasks done": I'm also fine to move those changes out to another PR if this eases the review process.

Yes, I think that would help with reviewing the docs changes. Thanks for looking into that!

internal/documentation/docs/pages/Builder.md
maxreichmann: Done in 1e086f5fa1d437427da994d18f2531681f4ff899

maxreichmann: Done in 1e086f5fa1d437427da994d18f2531681f4ff899

packages/project/lib/graph/ProjectGraph.js
maxreichmann: > This means that the BuildServer integration tests are likely unaffected by this change, which is fine.

I understand why to move the logic to server.js (originally I also thought about this). But isn't it also fine in ProjectGraph.js? I mean, it'd be in the dedicated `serve()` function outside of `build()` and we would have more realistic tests in `BuildServer.integration.js`. With the other way, the tests are logically correct but tend to be more unit test like.

RandomByte: ProjectGraph's `serve` might become public API at some point (similar to the `build` method). In that case we should not apply `@ui5/server`-specific configuration, but rather keep it as clean and generic as possible so nobody gets suprised.

maxreichmann:
~~_Removing the Draft state of this PR to check merge conflicts_~~

~~❗️ Pls don't start a review yet :)~~

packages/project/lib/graph/ProjectGraph.js
maxreichmann: Done in 8827f4ea29b4803d30031b26944a2c1c5cdd7e58

maxreichmann:
_✅ Update: Rebased. CI problems now fixed. PR ready for review_

packages/server/lib/server.js
```diff
@@ -188,6 +188,16 @@ export async function serve(graph, {
...
 		initialBuildIncludedDependencies.push("sap.ui.core");
 	}
+
+	// Explicitly exclude task "generateVersionInfo" for Server builds
+	// because middleware "versionInfo" will generate the version info anyways.
+	if (!Array.isArray(excludedTasks)) {
+		excludedTasks = [];
+	}
+	if (!excludedTasks.includes("generateVersionInfo")) {
+		excludedTasks.push("generateVersionInfo");
```
RandomByte:
```suggestion
		excludedTasks = [...excludedTasks, "generateVersionInfo"];
```

So we don't alter the supplied array by reference

packages/builder/test/lib/builder/builder.js
```diff
@@ -71,14 +71,20 @@ async function checkFileContentsIgnoreLineFeeds(t, expectedFiles, expectedPath,
...
-					try {
-						t.deepEqual(JSON.parse(currentContent), JSON.parse(expectedContent), expectedFile);
-					} catch (e) {
-						t.falsy(e, expectedFile);
+			} else if (expectedFile.endsWith(".json")) {
+				try {
+					const currentJson = JSON.parse(currentContent);
+					const expectedJson = JSON.parse(expectedContent);
+					// Check if file is "sap-ui-version.json" and ignore the buildTimestamp property for comparison:
+					if (expectedFile.endsWith("sap-ui-version.json")) {
```
matz3: This is not a safe check.
Better check for `relativeFile === "sap-ui-version.json"` to be sure the file is exactly with that name.
The `expectedFile` variable contains the full path, which is why something like `endsWith` was needed. But right now, a file called `not-sap-ui-version.json` would also match here, which is unexpected.

Note: In this case it is not a severe problem, as this is just part of our own tests, but in other scenarios, such incomplete checks could cause more severe problems or even security vulnerabilities, so I would propose that we apply those standards to all of our code, to be consistent.
Also, CodeQL or similar tools might flag this code for the same reason.

matz3 (commented):
LGTM, just a minor comment.

Documentation changes look fine to me.

packages/builder/test/lib/builder/builder.js
maxreichmann: Done

maxreichmann:
_Rebased and test problem fixed. Review ready again_

packages/builder/test/lib/builder/builder.js
```diff
@@ -67,18 +67,22 @@ async function checkFileContentsIgnoreLineFeeds(t, expectedFiles, expectedPath,
...
-				expectedContent = JSON.parse(expectedContent.replace(/(:\s+)(\d+)/g, ": 0"));
-				t.deepEqual(currentContent, expectedContent);
-			} else {
-				if (expectedFile.endsWith(".json")) {
+			if (expectedFile.endsWith(".json")) {
+				const currentJson = JSON.parse(currentContent);
+				const expectedJson = JSON.parse(expectedContent);
+				if (relativeFile === path.join("resources", "sap-ui-version.json")) {
+					// Check if file is "sap-ui-version.json" and ignore the buildTimestamp property for comparison:
 					try {
```
matz3: I assume the `try` / `catch` existed for potential parsing errors in `JSON.parse`. So parsing should be done within the `try` block to make sure errors are properly reported as assertions.

maxreichmann: Done

package-lock.json
```diff
@@ -2671,9 +2671,6 @@
 				"arm64"
 			],
 			"dev": true,
-			"libc": [
```
RandomByte, line 2674: package.json was not modified, so this change might not be intended?

packages/project/test/lib/build/BuildServer.integration.js
```diff
@@ -1333,6 +1441,18 @@ class FixtureTester {
 		const expectedProjects = Object.keys(projects);
 		this._t.deepEqual(projectsInOrder, expectedProjects);
 
+		// Optional check: Assert executed tasks
+		for (const [projectName, expected] of Object.entries(projects)) {
+			if (!expected.executedTasks) {
+				continue; // no executedTasks specified -> skip the check
+			}
+			const expectedExecuted = expected.executedTasks || [];
```
RandomByte: Already checked above?

```suggestion
			const expectedExecuted = expected.executedTasks;
```

packages/server/lib/server.js
```diff
@@ -188,6 +188,16 @@ export async function serve(graph, {
...
 	}
+
+	// Explicitly exclude task "generateVersionInfo" for Server builds
+	// because middleware "versionInfo" will generate the version info anyways.
+	if (!Array.isArray(excludedTasks)) {
+		excludedTasks = [];
+	}
+	if (!excludedTasks.includes("generateVersionInfo")) {
+		excludedTasks = [...excludedTasks, "generateVersionInfo"];
+	}
```
RandomByte, line 199: Removing these lines only causes failure of a single unit test that is actually for the serveIndex middleware.

I see that server.js only has very basic unit tests as of today, but I think at last stubbing and validating the arguments provided to `graph.serve` would be good here.

RandomByte (changes requested):

package-lock.json
matz3: This happens when running `npm install` with npm v10.x or maybe also with an older v11.x version.
Please use the latest npm v11.x instead (currently `11.18.0`).

maxreichmann: Done

packages/project/test/lib/build/BuildServer.integration.js
maxreichmann: Done

packages/server/lib/server.js
maxreichmann: I've now added 3 tests inside `packages/server/test/lib/server/server.js` verifying that the `excludedTasks` option is correctly edited and passed over to `graph.serve()`.

matz3:
LGTM

maxreichmann:
Kindly asking @KlattG now for reviewing the docu changes under `internal/documentation/docs` :)

internal/documentation/docs/pages/Server.md
```diff
@@ -145,6 +145,10 @@ In case a directory has been requested, this middleware renders an HTML with a l
 ## Standard Tasks
 As with the UI5 Builder, a set of standard tasks is being executed during a server build. Individual tasks can be included or excluded using the respective CLI options `--include-task` and `--exclude-task`.
 
+::: info Exception
+For the server, task [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo.md) is not executed because the corresponding middleware [`versionInfo`](#versioninfo) creates the same output file `sap-ui-version.json`.
```
KlattG:
```suggestion
For the server, the [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo.md) task is not executed, because the corresponding [`versionInfo`](#versioninfo) middleware creates an identical `sap-ui-version.json` output file.
```

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -320,6 +320,16 @@ Update your `ui5.yaml` configuration to reference an existing middleware instead
...
 
+## `sap-ui-version.json`
+
+When running `ui5 build`, the standard task [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) (producing a `sap-ui-version.json` file under `resources/`) is now **executed by default**. This applies to every build type (default, jsdoc, and self-contained) for projects of type `application`. For projects of other types (e.g. `library`), the behavior remains the same and [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) is not executed.
+
+::: info Tip
+To see which standard tasks are executed for each project type, check out the [Standard Tasks](../pages/Builder#standard-tasks) table in the UI5 Builder page.
+:::
+
+When running `ui5 serve`, [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) continues to **not be executed by default** because a corresponding middleware [`versionInfo`](../pages/Server.md#versioninfo) produces the same `sap-ui-version.json` output file.
```
```suggestion
When running `ui5 serve`, [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) continues to 
not run by default. The [`versionInfo`](../pages/Server.md#versioninfo) middleware creates an identical `sap-ui-version.json` output file.
```

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -320,6 +320,16 @@ Update your `ui5.yaml` configuration to reference an existing middleware instead
 | `serveThemes`      | CSS files pre-built by `buildThemes` task and served via `serveResources` | `serveResources` |
 | `testRunner`       | TestRunner resources served via `serveResources` from the UI5 framework | `serveResources` |
 
+## `sap-ui-version.json`
+
+When running `ui5 build`, the standard task [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) (producing a `sap-ui-version.json` file under `resources/`) is now **executed by default**. This applies to every build type (default, jsdoc, and self-contained) for projects of type `application`. For projects of other types (e.g. `library`), the behavior remains the same and [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) is not executed.
```
```suggestion
When you run `ui5 build`, the standard task [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) now runs by default. This task generates an `sap-ui-version.json` file in the `resources/` directory. The task runs for all build types (default, jsdoc, and self-contained) in projects of type `application`. For other project types (such as `library`), the behavior remains unchanged: [`generateVersionInfo`](../api/module-@ui5_builder_tasks_generateVersionInfo) does not run.
```

KlattG (changes requested):
See my comments below

KlattG (dismissed):
LGTM

RandomByte (dismissed):

RandomByte (approved):

PR #1394, opened 2026-05-22
[PoC] feat: Clean cache command

d3xter666:
JIRA: CPOUI5FOUNDATION-891

Summary of the implementation & challenges: [CACHE_CLEAN_SUMAMRY.md](https://github.com/UI5/cli/blob/9b00241e1cfbb742319cb15d2234a86f0dd8ace3/CACHE_CLEAN_SUMAMRY.md)

PR #1455, opened 2026-07-17
feat(cli): Add ui5 cache clean command

d3xter666:
This PR adds a new ui5 cache clean command to provide a safe, first-class cache cleanup workflow.

**Why**
Manual cache directory deletion is error-prone and inconvenient for CI/scripted usage.

**What changed**
- Added ui5 cache clean with interactive confirmation by default.
- Added non-interactive mode via --force.
- Added detailed grouped preview/result output via --verbose.
- Added stale cache recovery for:
    - interrupted framework cleanup directories
    - deferred build-cache vacuum cleanup

**Output behavior**
- Non-verbose: silent when nothing to clean; warning + confirmation when cache is found; fully silent with --force.
- Verbose: shows cache path, grouped Active/Stale preview, and grouped cleanup result.

JIRA: CPOUI5FOUNDATION-891


internal/documentation/docs/pages/Troubleshooting.md
```diff
@@ -20,28 +20,36 @@ UI5 CLI stores several kinds of data under your user's home directory in `~/.ui5
...
+
+Any required framework dependencies will be re-downloaded during the next UI5 CLI invocation.
+
 ::: info
-If you have configured a custom data directory via `UI5_DATA_DIR` or `ui5 config set ui5DataDir`, replace `~/.ui5/` with that path. See [Changing UI5 CLI's Data Directory](#changing-ui5-cli-s-data-directory).
+If you have configured a custom data directory via `UI5_DATA_DIR` or `ui5 config set ui5DataDir`, the `ui5 cache clean` command will clean up that location instead of the default `~/.ui5/`. See [Changing UI5 CLI's Data Directory](#changing-ui5-cli-s-data-directory).
+:::
+
+::: warning
+Only remove these directories, or run `ui5 cache clean`, when no UI5 CLI process and no `@ui5/*` API consumer is actively running. Running `ui5 cache clean` while `ui5 build` or `ui5 serve` is in progress can break the running process and lead to failed or inconsistent results.
```
matz3, line 59: Shouldn't this warning also be printed by the command when the user is asked for confirmation?

d3xter666: And what about if the user proivdes the `--yes` flag (`ui5 cache clean --yes`) ?
Then the command will be executed regardless if there are already running concurent builds / serve.

Maybe, we can include this information also in `ui5 cache --help`. What do you think?

matz3: Yes, that would also be helpful. For `--yes` I guess there's no point in showing it.

d3xter666: Done in https://github.com/UI5/cli/pull/1455/changes/0e631fa761826e9d7415eee6594de25f0b1fc07d

matz3:
Note: When I run `npm i` with npm v11.18.0 I get a diff in the `package-lock.json`, adding `yeso` to the `@ui5/cli` deps.
Looks like the lockfile is not up to date?

packages/builder/lib/lbt/resources/ResourceCollector.js
```diff
@@ -55,11 +55,11 @@ class ResourceCollector {
 	}
 
 	/**
-	 * Comma separated list of components to which orphaned resources should be added.
```
matz3, line 58: I don't think we should rename this term here. The class has a `createOrphanFilters`, so I would propose that we do not change this.

packages/project/lib/build/cache/BuildCacheStorage.js
```diff
@@ -511,6 +516,88 @@ export default class BuildCacheStorage {
...
+	 * DROP TABLE never reads row data; it only removes the schema entry and adds
+	 * pages to the freelist. Call {@link vacuum} afterwards to reclaim disk space.
+	 *
+	 * A persistent marker is set so that a deferred VACUUM can be detected on the
+	 * next invocation even if the process exits before {@link vacuum} runs.
+	 *
+	 * @returns {number} Database size in bytes before the drop (pending reclamation after vacuum)
+	 */
+	dropAllRecords() {
+		const tables = ["content", "index_cache", "stage_metadata", "task_metadata", "result_metadata"];
```
matz3: Same array values as above, so they need to stay in sync. Why not extracting this into a constant for all table names? Maybe we need it also somewhere else in future.

packages/project/lib/build/cache/BuildCacheStorage.js
```diff
@@ -511,6 +516,88 @@ export default class BuildCacheStorage {
...
+	 * Runs VACUUM to reclaim disk space from freed pages and clears the pending marker.
+	 * This is the slow half of the two-phase cache clean; call it from
+	 * <code>cleanAdditional</code> after the fast {@link dropAllRecords} pass.
+	 *
+	 * @returns {number} Number of bytes freed
+	 */
+	vacuum() {
+		const bytesBefore = this.getDatabaseSize();
+		this.#db.exec("VACUUM");
+		this.#db.exec("UPDATE _vacuum_pending SET pending = 0 WHERE rowid = 1");
```
matz3, line 598: Shouldn't this also be an transaction?

packages/project/lib/build/cache/CacheManager.js
```diff
@@ -337,4 +338,144 @@ export default class CacheManager {
...
+
+	/**
+	 * Get build cache info for the current version.
+	 *
+	 * @public
+	 * @static
+	 * @param {string} ui5DataDir Resolved absolute path to UI5 data directory
+	 * @returns {Promise<{path: string, size: number}|null>} Build cache info or null
+	 */
+	static async getCacheInfo(ui5DataDir) {
```
matz3: This and the next three methods are almost identical, and it doesn't look like it would be too complex to refactor them into a common method that can be used for the individual ones, without creating more complexity.

packages/project/lib/ui5Framework/cache.js
```diff
@@ -0,0 +1,241 @@
...
+ */
+
+const FRAMEWORK_DIR_NAME = "framework";
+
+/**
+ * Prefix used for staging directories created during an atomic framework cache clean.
+ * The directory is renamed to this prefix + a random hex suffix before deletion so that
+ * the original path is immediately removed and the deletion can proceed outside the rename.
+ */
+const STAGING_DIR_PREFIX = "_framework_to_delete_";
```
matz3: We should use a different term than `staging` here as there is already a `staging` dir for downloading the framework libraries (`.ui5/framework/staging`), which can be confusing as they are fundamentally different. At least an addition, like "removal" or "delete" or something would help to identify that they are not the same.

packages/project/lib/build/cache/BuildCacheStorage.js
d3xter666: Unfortunately, it's not possible:
https://www.sqlite.org/lang_vacuum.html
```
A VACUUM will fail if there is an open transaction on the database connection that is attempting to run the VACUUM
```

The order of the queries should be safe enough. Even if someting breaks the `VACUUM`, it will re-run on the next `cache clean` command as the flag `_vacuum_pending` is still present. Even if the `VACUUM` is successful, it shouldn't be an issue running it in the next `clean cache` iteration- it will be fast and then the flag will be updated.

matz3: Understood, thanks!

packages/project/lib/build/cache/BuildCacheStorage.js
d3xter666: Done in https://github.com/UI5/cli/pull/1455/commits/af2ec4002dff0fa72ae222c4181a46ca4102616a

packages/project/lib/build/cache/CacheManager.js
d3xter666: Done with https://github.com/UI5/cli/pull/1455/commits/144ba926b058ea861421094f295adc3019d9c695

packages/project/lib/ui5Framework/cache.js
d3xter666: I refactored it. Couldn't find a great name, but hopefully this will work now: https://github.com/UI5/cli/pull/1455/commits/b6150af7e2479fa55b24e632752e2af973e90943

packages/builder/lib/lbt/resources/ResourceCollector.js
d3xter666: It is by accident. I did a mass "orphaned" to "stale". I will double check on the rest of the items

packages/project/lib/ui5Framework/cache.js
matz3: I like it 👍🏻

packages/project/package.json
```diff
@@ -18,6 +18,8 @@
 	],
 	"type": "module",
 	"exports": {
+		"./internal/cache/CacheManager": "./lib/build/cache/CacheManager.js",
```
matz3: Why omitting the `build` part in the exported module name here? I know this is internal anyways, but this is the build cache manager, not the general cache manager of this package, so the `build` makes sense to me to keep.

d3xter666: Copy / paste error! Thanks!
Editted it with https://github.com/UI5/cli/pull/1455/commits/eb5321d1324815c3d326837be55cc49862d3b657

packages/cli/lib/cli/commands/cache.js
```diff
@@ -0,0 +1,165 @@
...
+	]);
+
+	const [additionalFrameworkResult, additionalBuildResult] = await Promise.all([
+		FrameworkCache.cleanAdditional(ui5DataDir),
+		CacheManager.cleanAdditional(ui5DataDir),
+	]);
+
+	if (isVerbose) {
+		const cleanedStaleFramework = staleInfo.length > 0 ? withAbsPath(additionalFrameworkResult, ui5DataDir) : [];
+		const cleanedStaleBuild = buildStaleInfo.length > 0 ? withAbsPath(additionalBuildResult, ui5DataDir) : [];
```
matz3: Re-using the initially gathered information after the confirmation means that the information might itself be `stale` (i.e. outdated).

This can be reproduced by running the command (without force) with one of the caches are filled (e.g. only framework), then filling the other cache (e.g. build cache), and afterwards confirming the question to clear the cache.
I got this output (note the "Removed null"):

```
❯ ui5 cache clean --verbose
verb cli:middlewares:base using @ui5/cli version 5.0.0-alpha.6 (from /github/UI5/cli/packages/cli/bin/ui5.cjs)
verb cli:middlewares:base using node version v24.18.0
Checking cache at /.ui5 …
verb build:cache:BuildCacheStorage Opening build cache database: /.ui5/buildCache/v0_7/cache.db
verb build:cache:BuildCacheStorage Opening build cache database: /.ui5/buildCache/v0_7/cache.db

The following cached data will be removed:

  Active Cache
    Framework
      • /.ui5/framework · 1 version of 6 libraries

Warning: Only run ui5 cache clean when no UI5 CLI process and no @ui5/* API consumer is actively running.
Running ui5 cache clean while ui5 build or ui5 serve is in progress can break the running process and lead to failed or inconsistent results.

Proceed with cache cleanup? (y/N) y
verb build:cache:BuildCacheStorage Opening build cache database: /.ui5/buildCache/v0_7/cache.db
verb build:cache:BuildCacheStorage Opening build cache database: /.ui5/buildCache/v0_7/cache.db

Cleanup result:

  Active Cache
    Framework
      ✓ Removed /.ui5/framework · 1 version of 6 libraries
    Build
      ✓ Removed null

Success: Cleaned Active Cache (Framework and Build)
```

d3xter666: Altough, this is a scenario which must not occur due to the warning we provide, I have added a guard about it and after accepting the prompt, the correct summary appears.

This issue also got me to think of other cases in this category. And I found this one, which is also fixed with the change:

- Terminal 1: Run `ui5 cache clean --verbose` and do not accept the prompt yet
- Terminal 2: Run `ui5 cache clean --force`
- Terminal 1: Accept the prompt now.
**Result**: The cache has been actually cleaned during the Terminal 2 execution and we need to provide the correct information, that there is not anything left to be cleaned.

Both cases rely now on the most recent information of the cache, derived after the prompt is accepted.
Fix: https://github.com/UI5/cli/pull/1455/commits/af83085307a6970bd2a6de7b74e764e6efbc3749

packages/cli/lib/cli/commands/helpers/cacheOutput.js
```diff
@@ -0,0 +1,239 @@
...
+				const detail = buildEntry.size > 0 ? `freed ${formatSize(buildEntry.size)}` : "";
+				items.push({absPath: buildEntry.absPath, detail});
+			}
+			staleCategories.push({title: GROUP_BUILD, items});
+		}
+		sections.push({title: SECTION_STALE_CACHE, categories: staleCategories});
+	}
+
+	if (sections.length === 0) {
+		process.stderr.write(`\n${chalk.italic(PARALLEL_CLEANUP_NOTICE)}\n\n`);
```
matz3: Why are here, and in other places two additional new-lines, and in case of a success only one? I wonder why two should be needed, but I think this should be consistent.

packages/cli/lib/cli/commands/cache.js
matz3: I'm not sure whether we can say it must not occur. The warning is only shown when the user first runs the command, so it could be the case that the user still has some process running, which populates one of the caches after the clean command was triggered. The user then exists the process as suggested by the warning and continues with the cache clean.

The commit solved the described behavior, but the `displayCacheInfo` function still receives the `staleInfo` / `buildAdditionalInfo`, based on the stale check that happened before the confirmation.
Now there is still the issue that the output claims it removed the stale build cache, but in fact it has been cleaned already. This is not known because the code relies on the initial stale cache information.

I would assume the easy fix is to call `CacheManager.getAdditionalCacheInfo()` again after the confirmation, as otherwise we do not know whether we needed to vacuum already, or only after the cleaning of the build cache.
Alternative would be to just include the vacuum into the normal cleanup (which would align it to how the framework cache does it, where it does rename + rmrf).

Yes, this might be nit-picking for this use-case, but to me this boils down to writing clean code, where we should not use outdated information at places where it is not supposed to be used.

packages/cli/lib/cli/commands/helpers/cacheOutput.js
d3xter666: Done

packages/cli/lib/cli/commands/cache.js
d3xter666: I have refactored it to always return the latest valid data. Now whatever case we're in, it will display only the really deleted information

matz3 (commented):
Fine for me from technical pov 👍🏻

internal/documentation/docs/pages/Troubleshooting.md
```diff
@@ -20,28 +20,43 @@ UI5 CLI stores several kinds of data under your user's home directory in `~/.ui5
...
 | `~/.ui5/server/` | Locally generated SSL certificate and private key for HTTPS / HTTP/2 mode | Yes — regenerated on next HTTPS server start; the new certificate must be re-trusted |
 
-::: warning
-Only remove these directories when no UI5 CLI process and no `@ui5/*` API consumer is actively running. Deleting files that are in use can cause running builds or servers to fail or produce inconsistent results.
-:::
-
 #### Resolution
 
-To free disk space, remove the relevant subdirectory.
+Use the dedicated cache clean command, which removes all cached data:
```
flovogt, line 25: @KlattG could you please do a review here?

package-lock.json
```diff
@@ -1262,20 +1262,30 @@
 			}
 		},
 		"node_modules/@emnapi/core": {
-			"version": "1.11.1",
-			"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.11.1.tgz",
-			"integrity": "sha512-RSvbQmHzdKzNsLYa/wHrbc3KN4sYLKAdPZxqiM2HATqv/SBk2/ENSHpvXGaLOMcsAyz0poEGqkmmKYG3OWiJEQ==",
+			"version": "1.11.3",
```
flovogt: @d3xter666 lockfile changes look weird. Could you please use npm v11 and see if the lockfile is still in this state?

d3xter666: Thanks! It used to have some problems with `npm ci` that's why I reset it locally. Now I have rabsed and aligned it with the main since we do not have here any related changes that will modify the `package-lock.json`

internal/documentation/docs/pages/Troubleshooting.md
```diff
@@ -20,28 +20,43 @@ UI5 CLI stores several kinds of data under your user's home directory in `~/.ui5
...
 ```sh
-rm -rf ~/.ui5/buildCache/
+ui5 cache clean --force
 ```
 
+The command removes the following cached data:
+- **UI5 Framework packages** — downloaded UI5 library files (`~/.ui5/framework/`)
+- **Build cache** — build data (`~/.ui5/buildCache/`)
+
+If a previous `ui5 cache clean` was interrupted (e.g. process killed or system crash), the command also detects and removes any leftover data from that interrupted operation, listed as separate entries:
```
KlattG:
```suggestion
If a previous `ui5 cache clean` was interrupted (for example, because the process was killed or the system crashed), the command also detects and removes any leftover data from that interrupted operation. This data is listed as separate entries:
```

KlattG (changes requested):

packages/project/lib/build/cache/BuildCacheStorage.js
```diff
@@ -9,6 +9,9 @@ const log = getLogger("build:cache:BuildCacheStorage");
 const METADATA_COMPRESSION_THRESHOLD = 4096;
 const CONTENT_COMPRESSION_THRESHOLD = 128;
 
+/** All live data table names */
+const DATA_TABLES = ["content", "index_cache", "stage_metadata", "task_metadata", "result_metadata"];
```
RandomByte, line 13: For the future (not for this PR), determining the tables dynamically would be a nice improvement.

_(see also https://github.com/UI5/cli/pull/1394#discussion_r3519050787)_

RandomByte (commented):
LGTM. Tested locally and works as expected.

My approval is pending UA review

KlattG (approved):
LGTM

PR #1475, opened 2026-07-23
refactor(logger): Restore cursor on ui5 serve termination

RandomByte:
The log-update package hides the terminal cursor on every render and only shows it again in its own `done()`, which InteractiveConsole calls from `disable()`. On CTRL+C the process is signal-terminated without `disable()` ever running, **so the cursor stays hidden in the user's shell**. The restore-cursor safety net (a bare `\x1b[?25h` with no newline) is also swallowed: the stderr write interceptor buffers non-newline-terminated writes into `partial` and only flushes them in `disable()`.

Register a handler for SIGINT/SIGTERM/SIGHUP/SIGBREAK (and an `exit` catch-all) in enable(), removed in disable(). Each signal handler tears the live region down via disable(), then re-raises the signal with process.kill(process.pid, signal) rather than calling process.exit(). Re-raising lets the cooperating handlers in profile.js and ProjectBuilder still run their async cleanup, and lets the default disposition set the 128+n exit code. `disable()` removes the handlers before re-raise so the re-sent signal reaches the default disposition instead of looping. 

`disable()` now also writes the show-cursor escape explicitly to its own stream, after uninstalling the interceptor. log-update's done() targets the real process.stderr and is gated on isTTY, so it can be a no-op; the explicit write restores the cursor regardless and is observable in tests.

### DIFF
--- a/packages/logger/lib/writers/InteractiveConsole.js
+++ b/packages/logger/lib/writers/InteractiveConsole.js
@@ -177,6 +177,11 @@ class InteractiveConsole {
 	// re-raise so cooperating handlers (profile.js, ProjectBuilder) still run.
 
 	#registerSignalHandlers() {
+		// Re-registering would strand the previous handlers on `process`:
+		// #deregisterSignalHandlers() can only detach what the map tracks.
+		if (this.#signalHandlers.size > 0) {
+			return;
+		}
 		for (const signal of TERMINATION_SIGNALS) {
 			const handler = () => this.#handleTerminationSignal(signal);
 			this.#signalHandlers.set(signal, handler);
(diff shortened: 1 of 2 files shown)


packages/logger/lib/writers/InteractiveConsole.js
```diff
@@ -148,9 +164,48 @@ class InteractiveConsole {
...
+			const handler = () => this.#handleTerminationSignal(signal);
+			this.#signalHandlers.set(signal, handler);
+			process.on(signal, handler);
+		}
+		// `exit` covers cooperating handlers that call process.exit(); it never
+		// fires on a bare signal kill, hence the signal handlers above.
+		const onExit = () => this.disable();
+		this.#signalHandlers.set("exit", onExit);
+		process.on("exit", onExit);
+	}
```
d3xter666, line 195: This method is not idempotent. Calling it twice, will just override `this.#signalHandlers`, and the firstly registered handlers will not be able to be deleted.
A check for registered signal in `this.#signalHandlers` will solve the issue

d3xter666 (changes requested):

packages/logger/lib/writers/InteractiveConsole.js
RandomByte: Good point! Fixed 👍

d3xter666 (approved):

PR #1476, opened 2026-07-24
refactor(builder): Run generateVersionInfo by default

maxreichmann:
⚠️ TESTING ONLY:

This branch is a copy of branch "run-generateVersionInfo-by-default" (PR #1453). It was created due to some CI problems in the original branch.


maxreichmann:
✅ _Closing this test branch because the original is now fixed_

PR #1478, opened 2026-07-24
refactor: Handle project definition changes

RandomByte:
## Problem

`ui5 serve` could not safely react to project-definition changes while running. Changes to files like `ui5.yaml`, `package.json`, workspace config, or a static dependency definition need a fresh project graph, but the server only watched source files and served an app it had already built. Switching branches or editing config could leave the running server pointing at an out-of-date graph, source paths that had moved, or short-lived watcher/build errors.

## Solution

This adds a way to handle definition changes, alongside the existing handling of source changes.

- Add `ProjectDefinitionWatcher` to watch the project-definition files.
- Add `Supervisor` to keep the HTTP server and port stable while it rebuilds the graph, Express app, and `BuildServer` behind them.
- Build the new server before tearing down the old one, so a failed rebuild leaves the last working server running.
- Recover a broken server by retrying until the new graph is stable: a fast burst for a mid-checkout race (sources still being written) and an indefinite slow poll for a dependency installed after the edit, instead of serving a half-restored project or giving up.
- Stop answering requests as soon as a definition change starts, so requests fail fast instead of hanging until the checkout finishes writing files.
- Stop the outgoing server's build loop at the same point, so it does not keep rebuilding against the shared build cache while the new server builds.
- Re-point live reload and the file watchers at the new graph after a successful swap.
- Make source watching survive branch switches and moved paths.
- Update the interactive console while the project is being re-resolved, including an error status line that stays until the server recovers.
- Add `serveMiddleware` for mounting UI5 middleware into another server.

## ProjectDefinitionWatcher and Supervisor

`ProjectDefinitionWatcher` watches the files that define the project graph: project `package.json` files, `ui5.yaml` or a custom `--config`, workspace config, and static dependency-definition files. It fires an early `definitionChanging` signal (for UI feedback) and, once the changes settle, a `definitionChanged` signal that tells the server to reinitialize. Settling matters for bursts of changes such as branch switches, where many files change at once.

The watcher belongs to `Supervisor`, not the `BuildServer`, so it stays alive across serving-stack swaps. After a successful reinitialization it is re-pointed at the new graph, while the source watcher keeps handling incremental rebuilds for the graph that is currently live.

```
  CLI
   |
   v
serve()  [server.js]
   |
   v
Supervisor  (owns the stable HTTP socket)
   |
   +-- binds port once; request dispatcher routes every request to #stack.app
   |
   +-- ProjectDefinitionWatcher  (@ui5/project)
   |    watches ui5.yaml, package.json, workspace config
   |    definitionChanging --> blank version in console; stop answering requests; stop the build loop
   |    definitionChanged  --> Supervisor.reinitialize()
   |
   +-- #stack  { app, BuildServer }   <-- swapped on reinit; dispatcher retargets silently
   |    |
   |    +-- BuildServer  (@ui5/project)
   |         watches source files; emits sourcesChanged
   |
   +-- #sourcesChangedRelay  (stable; live-reload clients stay connected across swaps)
```

`Supervisor` keeps the socket and the relay for the whole run. Everything else -- the serving app, the `BuildServer`, and the `ProjectDefinitionWatcher` -- is rebuilt on a definition change and swapped in behind the request dispatcher, without dropping the port or disconnecting open browsers.

## Swap lifecycle

Reinitialization follows a small state machine. State only changes through `#setState`, which checks each change against a table of allowed transitions:

```
HEALTHY    --(definition change)--------> RESOLVING
RESOLVING  --(resolve+build ok)---------> HEALTHY
RESOLVING  --(resolve/build failed)-----> DEGRADED
DEGRADED   --(scheduled recovery)-------> RECOVERING
RECOVERING --(converged+built)----------> HEALTHY
RECOVERING --(still failing)------------> DEGRADED
any        --(destroy)------------------> DESTROYED  (terminal)
```

The state decides what is allowed: whether a swap is already running, whether the server has been shut down, and which transitions are legal. A `HEALTHY` swap resolves the graph once; a `DEGRADED` swap runs the recovery loop described below. If several `definitionChanged` events arrive while a swap is running, they collapse into one final pass, so a burst of edits only resolves once against the settled files.

One switch controls whether the server answers requests. When `definitionChanging` fires, the current `BuildServer` stops serving readers: it rejects the requests already waiting and fast-fails new ones. Serving is turned back on in a single place, used by both swap outcomes (success and failure), so no code path can accidentally leave the server stuck not answering.

`definitionChanging` also stops the current `BuildServer`'s build loop. Rejecting readers alone leaves it running: the checkout's source writes keep aborting and restarting builds on the outgoing server, which then runs at the same time as the new server's first build against the same build cache (a single refcounted handle shared by both). That contention corrupts the cache and interleaves the two builds' progress on the shared console feed (the "5/1" counter). Stopping the loop aborts the in-flight build, cancels the pending and deferred restarts, and blocks new ones, while already-built resources keep serving. Unlike the reader switch, this one is one-way: it is never turned back on. A clean swap destroys the old server, and a failed re-resolve keeps it serving already-built resources without rebuilding off the wrong branch's sources, so the loop stays off until teardown either way.

## Degraded recovery

If a re-resolve fails, the previous server keeps running, but its graph no longer matches the files on disk. The server is marked degraded: a middleware step sends every request to the error page (not just page loads), and the interactive console shows an error status line that stays put. The next successful swap, on a valid edit, clears the mark.

A degraded server also schedules its own recovery, on a timer, in two phases. Within a fast `RecoveryBudget` (5 attempts, ~550 ms apart) it retries quickly; once that budget is spent it starts a slow poll (`SLOW_RECOVERY_INTERVAL_MS`, 30 s) that repeats indefinitely. The two phases cover two different failure shapes.

The fast phase covers a transient checkout race. A branch switch writes `package.json`, `ui5.yaml`, and the source files as one operation, and on a big tree (or a slow disk, or Git LFS fetching files) those arrive in several waves, so a quick retry lands once the tree finishes settling.

The slow phase covers a dependency declared in `package.json` but installed minutes later (a plain `npm install` after the edit). The `ProjectDefinitionWatcher` only watches projects in the last resolved graph, so a brand-new `node_modules/<dep>` fires no file event; only a timer notices it. The slow poll never gives up on its own -- it stops on a successful swap, `destroy()`, or the next `definitionChanging` (a real edit, which also restores a full fast budget).

Each recovery swap resolves against a tree that may still be settling, so a single resolve can land in the gap between waves and see a tree still missing files, or missing a symlink target that a restored `package.json` points at. So recovery also loops within the swap (`#convergeRecoveryGraph`): resolve, list the project roots, and if that list changed from last time, wait for another quiet window and resolve again. It stops once two resolves in a row agree on the roots.

```
DEGRADED (last-good stack serving)
   |
   +-- schedule the next recovery on a timer
   |     within RecoveryBudget (5 attempts / 60s) --> fast, ~550 ms apart
   |     budget spent                             --> slow poll, 30 s, rescheduled indefinitely
   |
   v
#convergeRecoveryGraph()  loop, bounded by RECOVERY_MAX_ITERATIONS
   |
   +-- resolve graph via graphFactory()
   +-- collect project-root set
   +-- roots subset of previous?  --yes--> converged: build + swap --> HEALTHY
   |                              --no-->  feed [candidate, last-good] to
   |                                       projectGraphSettleWatcher, await quiet, re-resolve
   v
 (each failed swap schedules the next timer) --> stay DEGRADED, keep slow-polling until it
                                               resolves, destroy(), or a definition change
```

The check is "did the root set shrink or stay the same", not "is it identical", so a branch that removes a dependency still counts as settled. Each pass hands the settle watcher both the new graph and the last known-good one, so a project that only the target branch adds gets watched for quiet as soon as it first shows up in a resolve. `RecoveryBudget` counts an attempt when a fast recovery is scheduled, not when it succeeds, so a persistently broken branch drains the fast allowance and drops to the slow poll rather than retrying fast forever. A real edit (`definitionChanging`) cancels any pending retry and restores a full fast budget; a clean swap resets it too. The failure log is dampened to match the indefinite poll: the first failure of an episode logs at `error`, repeats at `verbose` while already degraded, so the slow poll does not print an error stack every interval.

## Middleware API

The server now keeps two jobs apart: building the middleware and owning the HTTP listener. That split is what makes the graph swap work: the HTTP server stays bound to its port while a new graph, middleware stack, and `BuildServer` are built and swapped in behind it. `serve()` still starts and owns the UI5 dev server, but under the hood it reuses smaller pieces for building the middleware, binding HTTP, and managing graph swaps.

It also adds a new `serveMiddleware()` API for outside consumers. It sets up the same UI5 readers/resources and the same standard and custom middleware chain that `ui5 serve` uses, but hands it back as connect/Express-compatible middleware instead of starting an HTTP server.

Other tools can now call `app.use(result.middleware)` on their own Express/connect server and stay in charge of the port, protocol, routing around UI5, and error handling. The returned `close()` function frees the `BuildServer`'s watcher and cache when the outside server shuts down.

```
Third-party server (e.g. Vite, custom Express app)
   |
   | const {middleware, close} = await serveMiddleware(graph, options)
   v
serveMiddleware()  [server.js re-export -> serveMiddleware.js]
   |
   +-- calls buildRouter()  [serve/stack.js]
   |    |
   |    +-- starts BuildServer  (@ui5/project)
   |    +-- assembles UI5 middleware chain onto an Express Router
   |    +-- returns { router, buildServer }
   |
   +-- returns { middleware: router, buildServer, close() }


Third-party app                       @ui5/server
+-------------------------+           +--------------------+
| const app = express()   |           |                    |
| app.use(middleware)  ---+---------> | UI5 router         |
| app.use(errorHandler)   |           | (build-on-request, |
| app.listen(8080)        |           |  index, CSP, ...)  |
|                         |           +--------------------+
| // teardown:            |
| listener.close()        |
| await close()        ---+---------> BuildServer.destroy()
+-------------------------+           (releases source watcher
                                       and build-cache handle)
```

The caller owns the port, the HTTP server, error handling, and the order of teardown. `serveMiddleware` never binds a socket, never attaches a live-reload WebSocket server, and never installs a final error handler -- those stay with the caller's server.

JIRA: CPOUI5FOUNDATION-1307


packages/project/lib/graph/ProjectDefinitionWatcher.js
```diff
@@ -0,0 +1,283 @@
...
+import parcelWatcher from "@parcel/watcher";
+import {getLogger} from "@ui5/logger";
+import {drainSubscriptions, WATCHER_BURST_SETTLE_MS} from "../build/helpers/watchUtil.js";
+import RecoveryBudget, {
+	WATCHER_RECOVERY_MAX_ATTEMPTS, WATCHER_RECOVERY_WINDOW_MS,
+} from "../build/helpers/RecoveryBudget.js";
+const log = getLogger("graph:ProjectDefinitionWatcher");
+
+// Default filename of the workspace configuration, resolved against cwd.
+const WORKSPACE_CONFIG_DEFAULT = "ui5-workspace.yaml";
```
matz3: We now have this default value defined in three places (`graph.js`, `graph/helpers/createWorkspace.js`, and here).
Shouldn't we rather clean it up and use the same constant in all places?

packages/project/lib/graph/ProjectDefinitionWatcher.js
```diff
@@ -0,0 +1,283 @@
...
+
+		// Static-graph mode: the dependency-definition file is a topology definition, so editing it
+		// changes the graph just like editing package.json does.
+		if (dependencyDefinitionPath) {
+			add(resolve(dependencyDefinitionPath));
+		}
+	}
+
+	#getIgnoreGlobs(dir) {
+		const segments = path.resolve(dir).split(/[\\/]+/);
```
matz3: Why not this?
```suggestion
		const segments = path.resolve(dir).split(path.sep);
```
`path.resolve` ensures that separators are according to the current system.

Code within `packages/project/lib/build/cache/` does the same in several places.

RandomByte: Suggestion applied and squashed into https://github.com/UI5/cli/pull/1478/commits/b8a676c3082f6872f6a30d0188032a3314cf8695

packages/project/lib/graph/ProjectDefinitionWatcher.js
RandomByte: Done in https://github.com/UI5/cli/pull/1478/commits/2e36412e7bd526a4812147809525f40e044ea47c

matz3 (approved):

PR #1488, opened 2026-07-28
refactor(builder): libraryLessGenerator: Rewrite svg.less imports

stopcoder:
The libraryLessGenerator processor already rewrites base.less imports from the UI5 resource path (sap/ui/core/themes/<theme>/base.less) to the theming-base-content path (Base/baseLib/<theme>/base.less) used by the Theme Designer. Extend this to also rewrite svg.less imports to Base/icons/<theme>/svg.less.

The Theme Designer resolves imports against the theming-base-content structure. When a customer creates a custom theme, icon overrides are stored under Base/icons/<custom_theme>/. The parent theme's library.less must import from the same Base/icons/ path so that custom content layered on top takes effect via LESS "last definition wins" semantics.

Extract a shared rewriteThemeDesignerImport() static method and drive both rewrites from a config array, so adding future rewrites is a one-liner.

JIRA: CPOUI5FRAMEWORK-812

Up-port of https://github.com/SAP/ui5-builder/pull/1223

### DIFF
--- a/packages/builder/lib/processors/libraryLessGenerator.js
+++ b/packages/builder/lib/processors/libraryLessGenerator.js
@@ -7,6 +7,7 @@ import Resource from "@ui5/fs/Resource";
 
 const IMPORT_PATTERN = /@import .*"(.*)";/g;
 const BASE_LESS_PATTERN = /^\/resources\/sap\/ui\/core\/themes\/([^/]+)\/base\.less$/;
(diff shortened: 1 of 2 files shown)


matz3:
Merging as `refactor`, not `feat`, as this is already part of v4, so there is no point in marking it as a new feature of v5.

matz3 (approved):

PR #1510, opened 2026-08-07
docs(documentation): Update server docs for v5 architecture

RandomByte:
Reflect the new UI5 Server architecture in the user-facing docs.

**Server.md:** The watch-mode section claimed config files are not monitored and that config changes require a restart. The Supervisor and ProjectDefinitionWatcher now re-resolve the graph and re-create the serving stack behind a stable HTTP server on a definition change. Split the section: keep source monitoring under Watch Mode Behavior, add a Project Definition Changes section covering ui5.yaml / package.json / workspace config / --config / --dependency-definition watching and the git checkout case. Requests are held with an auto-reloading page while the graph re-resolves. The restart caveat now scopes to custom task and middleware implementation code.

**index.md:** Add Node.js API examples for starting a server via serve() and for embedding the middleware via serveMiddleware().

JIRA: CPOUI5FOUNDATION-1307

### DIFF
--- a/internal/documentation/docs/index.md
+++ b/internal/documentation/docs/index.md
@@ -142,3 +142,97 @@ async function buildApp(projectPath, destinationPath) {
 }
 ```
 :::
+
+#### Starting a Server
+
+`@ui5/server` starts a development server for a project graph. It binds a port, serves the built resources, watches the sources, and rebuilds on demand.
+
+::: code-group
+```js [ESM]
+import {graphFromPackageDependencies} from "@ui5/project/graph";
+import {serve} from "@ui5/server";
+
+async function startServer(projectPath) {
+    const graph = await graphFromPackageDependencies({
+        cwd: projectPath
+    });
+    const {port, close} = await serve(graph, {
+        port: 8080,
+        changePortIfInUse: true
+    });
+    console.log(`Server started on port ${port}`);
+
+    // Later, to stop the server:
+    // await new Promise((resolve) => close(resolve));
+}
+```
+
+```js [CommonJS]
+async function startServer(projectPath) {
+    const {graphFromPackageDependencies} =
+        await import("@ui5/project/graph");
+    const {serve} = await import("@ui5/server");
+    const graph = await graphFromPackageDependencies({
+        cwd: projectPath
+    });
+    const {port, close} = await serve(graph, {
+        port: 8080,
+        changePortIfInUse: true
+    });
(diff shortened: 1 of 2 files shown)


internal/documentation/docs/index.md
```diff
@@ -141,4 +141,98 @@ async function buildApp(projectPath, destinationPath) {
...
+
+    // Later, to stop the server:
+    // await new Promise((resolve) => close(resolve));
+}
+```
+:::
+
+#### Embedding the Middleware
+
+`serveMiddleware` assembles the UI5 middleware as a single connect/Express handler, for mounting into an HTTP server you own instead of starting one. It does not bind a port or attach the Live Reload WebSocket server.
```
KlattG:
```suggestion
`serveMiddleware` assembles the UI5 middleware as a single connect/Express handler for mounting into an HTTP server you own instead of starting one. It does not bind a port or attach the LiveReload WebSocket server.
```

KlattG (changes requested):

internal/documentation/docs/index.md
RandomByte: We already have several docs writing it like "Live Reload" (also occasionally "live reload"). For instance: https://ui5.github.io/cli/next/pages/Server#livereload and https://ui5.github.io/cli/next/updates/migrate-v5#live-reload

What is your recommendation here?

KlattG: I'd recommend to use the correct spelling from now on and fix it elsewhere once you have some free capacity.

KlattG (changes requested):

internal/documentation/docs/index.md
RandomByte: Thanks, I will implement the changes.

Just to clarify: We are not using a tool called "LiveReload" (there are several tools with such a name). 

Instead, we have implemented our own functionality that automatically reloads the browser whenever something changes. Such a feature is often called "Live Reload" or "Hot Reload".

KlattG: Oh, I see. I did a simple search, found a "LiveReload" tool, and assumed that we were using this. If this is not the case, I retract my previous review. If we're talking about a tool of our own, we're of course free to call it and spell it any way we like - although I'd prefer to name it in a way that precludes confusion with other existing tools out there.

RandomByte: Thanks. I reverted the change

KlattG (approved):

PR #1511, opened 2026-08-07
fix(project): Add polling file watcher for containers

### LINKED ISSUE #1479: UI5 Build Cache does not work in case process is executed in a container
UI5 Build Cache does not work properly in case process is executed in a container. 

Initially reported by @henriquemattos.

## Steps to Reproduce the Issue
1. Execute `ui5 serve` in a project in a container (docker, podman, etc.)
2. Open the browser on your host and request UI5 project served by the container
3. Change any served file in the UI5 project 
4. Project is not rebuild and no live reload is triggered

## Context
 * UI5 Module Version (output of `ui5 --version` when using the CLI): **`@ui5/cli@v5.0.0-alpha.6`**
 * Node.js Version: **`24`**
 * npm Version: `11`

RandomByte:
## What

Add a polling file watcher for the incremental build. It is used automatically inside containers and can be forced anywhere with `UI5_WATCH_MODE`.

## Why

On a bind-mounted volume inside a container, Linux inotify reports the container's own writes but not writes made to the same volume from outside the container. Podman is the common case. The incremental build derives "what changed" solely from `@parcel/watcher` events, so those missed events break rebuilds and live reload, with no error to catch. Polling reads the tree directly, so it sees every change regardless of where it originated.

## How it works

A new module, `fileWatcher.js`, selects the backend and exposes a `subscribe()` matching `@parcel/watcher`'s exact contract, so a caller cannot tell which backend is active. All three watcher consumers (`WatchHandler`, `ProjectDefinitionWatcher`, `projectGraphSettleWatcher`) call it instead of `@parcel/watcher` directly. The native backend stays `@parcel/watcher`; the polling backend lives in a separate `pollingWatcher.js` that `fileWatcher.js` imports lazily, only when polling is selected, so the module stays off the common native path.

Both backends are imported lazily. `@parcel/watcher` resolves a native binding at load time and throws when the platform's prebuilt binary is not installed (for example on an unsupported platform or after an install with `--omit=optional`), so a static import would break every consumer before the polling fallback could run. Loading it on demand keeps that failure contained: when the native backend is selected but cannot load, `subscribe()` logs a warning and falls back to polling, which needs no native code.

The backend is chosen once per process and memoized:

- `UI5_WATCH_MODE=polling|native` forces it.
- Otherwise polling is the default inside a container, native elsewhere.
- A container is detected from the runtime marker files (`/.dockerenv`, `/run/.containerenv`) and PID 1's cgroup.

Container detection is a heuristic: it can miss an exotic sandbox or match a host whose PID 1 cgroup carries those names, so `UI5_WATCH_MODE` is the escape hatch either way.

The polling path walks the tree and diffs an mtimeMs+size snapshot every 250 ms (rescheduling after each walk so a slow crawl cannot overlap the next), emitting the same `{type, path}` events the native backend would. Errors flow through the callback so each consumer's existing recovery path fires unchanged. The 250 ms interval stays below `WATCHER_BURST_SETTLE_MS` (550 ms) so downstream event batching still holds. Ignore globs reuse micromatch and prune ignored directories so `node_modules` is never crawled.

## Configuration

- `UI5_WATCH_MODE=polling|native` forces the file watcher backend.

Documented under [Troubleshooting](https://ui5.github.io/cli/v5/pages/Troubleshooting/).

Fixes: #1479

### DIFF
--- a/internal/documentation/docs/pages/Troubleshooting.md
+++ b/internal/documentation/docs/pages/Troubleshooting.md
@@ -114,6 +114,18 @@ Cross Environment via [cross-env](https://www.npmjs.com/package/cross-env):
 cross-env UI5_CLI_NO_INTERACTIVE=1 ui5 serve
 ```
 
+### File Changes Not Picked up in a Container
+
+When `ui5 serve` runs a build internally, it watches the project's files and rebuilds on change. On a bind-mounted volume inside a container, file system events are often only reported for the container's own writes, but not for writes made to the same volume from outside the container. If you edit files on the host and no rebuild or live reload happens inside the container (commonly observed with Podman), the "native" file watcher of the UI5 CLI is not receiving those events.
+
+For this reason, UI5 CLI attempts to detect a container environment and automatically switch to a "polling" file watcher. That detection is a heuristic and can be wrong in some environments.
+
+If you encounter this problem in your container-based development setup, try setting the `UI5_WATCH_MODE` environment variable to `polling` to force the polling watcher, or to `native` to force the native watcher.
+
+::: info
+Polling reads the watched files on an interval, so it reports changes regardless of where they originate, at the cost of more CPU than the event-based native watcher. Use it only when the native watcher fails to detect your file changes.
+:::
+
 ### Changing UI5 CLI's Data Directory
 
 UI5 CLI's data directory is by default at `~/.ui5`. It's the place where the framework artifacts are stored.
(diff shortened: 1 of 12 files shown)


RandomByte:
@matz3 kindly re-test in podman. Hopefully the new environment detection will choose the polling backend automatically now.

matz3:
Successfully tested with Podman v6.0.2 and `node:22-alpine`.
With Docker v29.4.1, it also detects the container and switches to `polling`. Forcing `native` in Docker also works for me, which again proofs that this issue is only observed with Podman.

matz3 (approved):

PR #1513, opened 2026-08-07
fix(server): Sequence destroy() after native teardown

RandomByte:
Fixes a unit test failing flakily in CI, most often on Node 22/24 on Windows, with `test/lib/server/reinitialize.js exited with a non-zero exit code: 3221225477`. That code is `0xC0000005`, a native access violation on process teardown, not a test-assertion failure. The defect is in production teardown code, so the fix lands in `Supervisor.js`; the flaky reinitialize test is what surfaced it.

## Root cause

Supervisor.destroy() passed the caller's callback to http.Server.close(), firing it on socket close. The BuildServer's teardown (the `@parcel/watcher` unsubscribe and the node:sqlite handle, which maps a 256 MB region of the database into memory) runs after that call and was not sequenced ahead of the callback, so server.close(resolve) resolved with SQLite still mid-close. The next test.serial opening a fresh DatabaseSync, or the worker exiting, then races the finalizing handle and raises an access violation on Windows (0xC0000005), surfacing as `exited with a non-zero exit code: 3221225477` in the reinitialize suite.

## Fix

Make destroy() a plain promise: start the socket close, then await the definition watcher and BuildServer teardown, and await the socket last. The promise resolves only once every native handle is closed. server.js adapts this to the existing close(callback) API and also returns the promise when no callback is passed, so callers can await it; a teardown error is passed to the callback (and rejects the promise) instead of being swallowed.

### DIFF
--- a/packages/server/lib/serve/Supervisor.js
+++ b/packages/server/lib/serve/Supervisor.js
@@ -546,10 +546,9 @@ class Supervisor extends EventEmitter {
 	 * Stops the server: closes live-reload, the HTTP socket, and the current BuildServer. Teardown
 	 * is tolerant: the socket is closed even if the BuildServer's destroy rejects.
 	 *
-	 * @param {Function} [callback] Invoked once the HTTP server has closed
 	 * @returns {Promise<void>} Resolves once teardown completes
 	 */
-	async destroy(callback) {
+	async destroy() {
 		// Move to the terminal state synchronously, before the first await, so an in-flight #swap or a
 		// late definitionChanged sees DESTROYED at its next guard and adopts nothing.
 		this.#setState(STATE.DESTROYED);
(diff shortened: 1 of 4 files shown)


packages/server/lib/serve/Supervisor.js
```diff
@@ -560,8 +560,19 @@ class Supervisor extends EventEmitter {
 		this.#definitionWatcher = null;
 		this.#liveReloadHandle?.close();
 		this.#detachRelay();
-		this.#httpServer?.close(callback);
 		this.#clearRecoveryTimer();
+		// The callback fires only after every native handle is closed, not on socket close. The
+		// BuildServer's teardown closes the @parcel/watcher subscriptions and the node:sqlite handle
+		// (which maps a 256 MB region of the database into memory); a caller resuming while SQLite is
+		// mid-close raises an access violation on Windows (0xC0000005). So start the socket close here
+		// but await it last.
```
matz3: I think it's okay to explain the background in the commit message, but I would like to not add all the details to the code comment here. We should only generally describe that the callback should be called after everything has been destroyed, which is obviously what a caller would expect.

Keeping the lengthy comment also adds the risk of keeping stale information, e.g. in case we change something about the node:sqlite details or other things, which are not really important here, but should be kept in the commit message..

matz3 (changes requested):

packages/server/lib/serve/Supervisor.js
RandomByte: Good point, thanks! Done

matz3 (commented):
LGTM, but the commit message / PR description of the "Fix" need to be updated, as they still describe the firing of the callback

RandomByte:
> LGTM, but the commit message / PR description of the "Fix" need to be updated, as they still describe the firing of the callback

Thanks, I missed that. Updated the commit message and PR body

matz3 (approved):

PR #1522, opened 2026-08-13
deps!: Require Node.js ^22.22.2 || ^24.15.0 || >=26.0.0

maxreichmann:
This bumps the project's node engine requirements:
--> from **^22.20.0 || >=24.0.0**
--> to **^22.22.2 || ^24.15.0 || >=26.0.0**

Alignes with requirements used in `@npmcli` packages and npm 12.
Solves engine conflict in https://github.com/UI5/cli/pull/1494.
Adds new node version (26) to CI test matrix.
Updates documentation.


internal/documentation/docs/pages/GettingStarted.md
```diff
@@ -2,7 +2,7 @@
 ## Installing the UI5 CLI
 ### Requirements
 
-- [Node.js](https://nodejs.org/) version v20.11.0 and higher or v22.0.0 and higher (v21 is not supported)
+- [Node.js](https://nodejs.org/) version v22.22.2 and higher or v24.0.0 and higher (v23 is not supported)
```
matz3: Good catch. This was still outdated.

matz3 (changes requested):
Two remarks:
- Should we also bump the v24 requirement, as we now require a v22 version from March 2026, but still allow the initial v24 release from May 2025? To me this does not fit right now, and the v22 version might contain more features (back-ports), than the v24. Also, looking at https://github.com/UI5/cli/pull/1494#issuecomment-5239036871 it was stated that the `@npmcli/config` package needs `^22.22.2 || ^24.15.0`. But actually it needs `^22.22.2 || ^24.15.0 || >=26.0.0` (based on version `11.0.1`)
- Should we rather declare this as a breaking change, to make our requirement changes clear in the next alpha release notes?

matz3:
Third remark:

~Do you know why the CI does not fail? I would have expected that the check via https://github.com/UI5/cli/pull/1360 catches that.~
EDIT: Nevermind. The dependency doesn't get updated in this PR, so it would only fail there. All good here, no CI issue.

flovogt:
> Should we also bump the v24 requirement, as we now require a v22 version from March 2026, but still allow the initial v24 release from May 2025? To me this does not fit right now, and the v22 version might contain more features (back-ports), than the v24. Also, looking at https://github.com/UI5/cli/pull/1494#issuecomment-5239036871 it was stated that the @npmcli/config package needs ^22.22.2 || ^24.15.0. But actually it needs ^22.22.2 || ^24.15.0 || >=26.0.0 (based on version 11.0.1)

I agree. Let set the version to ^22.22.2 || ^24.15.0 || >=26.0.0 for now

> Should we rather declare this as a breaking change, to make our requirement changes clear in the next alpha release notes?

As we are still in the alpha phase every new release might be breaking. As we edit the changelog for initial v5 later anyway, I guess its fine to mark this as breaking

matz3:
> As we are still in the alpha phase every new release might be breaking. As we edit the changelog for initial v5 later anyway, I guess its fine to mark this as breaking

Yes, we need to state the breaking changes compared to v4 in the final v5.0.0 release. I agree that we should add this breaking change to inform front-runners who want to test our pre-releases.

EDIT: Sorry, I first misunderstood your point.

maxreichmann:
Update:
- Set node version requirements to: **^22.22.2 || ^24.15.0 || >=26.0.0**
- Add node 26 to CI test matrix (@UI5/ui5-team-cor-fnd ?)

package-lock.json
```diff
@@ -448,14 +452,14 @@
 			}
 		},
 		"node_modules/@babel/generator": {
-			"version": "7.29.7",
-			"resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.29.7.tgz",
-			"integrity": "sha512-DkXD5OJQaAQIdZ1bt3UZdEnHAn9Imd3IVBdX03UFe+ony9Ojw5pzr9YVKGDY1jt+Gcn/FnGkNf8r+Vj5NOJWtQ==",
+			"version": "7.29.8",
```
matz3: Is it intended that this PR also updates some of our dependencies? I would have expected to only see the `engines` fields changed in this file.

maxreichmann: Fixed

matz3:
`dep` is not a valid type (I assume you meant [deps](https://github.com/UI5/cli/blob/535e5ae0e027b400dc0ad966cd7212a0a4ce1017/release-please-config.json#L62)). Please also adjust the commit(s) so that the CI can validate the commit message.
Regarding the title, I would rather expect something that makes clear that we now **require** those versions for our packages. The current `Set project node version` can sound like we are using the versions now, but it does not state that consumers need to adjust, which is the case here.
I would suggest `deps!: Require Node.js ^22.22.2 || ^24.15.0 || >=26.0.0`, which fits to what most other npm packages describe in their changelogs.

matz3 (approved):

PR #1524, opened 2026-08-14
test(server): Speed up reinitializeMissingDependency via shortened poll

matz3:
The test sat through the real 30 s slow-recovery interval, taking 35 s+. Expose the interval through a test-only __internals__ handle on the Supervisor (NODE_ENV-gated, undefined in production) so the test can shorten it to 500 ms.

Rework the fast-budget drain to count re-resolve failures instead of detecting a quiet window, since a slow interval near the fast cadence no longer distinguishes the two phases by timing. The test now runs in ~6 s and still drives the genuine end-to-end recovery path.


RandomByte (approved):

matz3:
There is still a flaky test for `ProjectDefinitionWatcher`, which suffers from a similar issue on Windows like in #1513. I will follow-up on that next week.

PR #1528, opened 2026-08-18
fix(project): Cancel pending settle timer before watcher recovery

matz3:
On Windows, a settle timer callback firing into a closed ReadDirectoryChangesW handle after recovery causes an access violation (0xC0000005). Cancel any pending timer before tearing down the subscriptions; it is re-armed by the first event on the new set.

Extract the timer-cancel and subscription-drain logic shared by #recoverWatcher and destroy into #cancelSettleTimer and #drainSubscriptions helpers to remove the duplication.


packages/project/lib/graph/ProjectDefinitionWatcher.js
```diff
@@ -230,6 +230,15 @@ class ProjectDefinitionWatcher extends EventEmitter {
 		}
 
 		try {
+			// Cancel any pending settle timer before tearing down the subscriptions. On Windows,
+			// ReadDirectoryChangesW holds a raw HANDLE; a timer callback firing into a closed handle
+			// after recovery causes an access violation (0xC0000005). The timer is re-armed naturally
+			// by the first event that arrives on the new subscriptions.
```
d3xter666: Do we need all the resoning here. I think there's much detail that needs a deeper look to understand the root cause.
From my perspective, it's completely resonable and expected to clear timeouts before releasing the property. Otherwise, side effects (like the one above) can be expected

d3xter666 (dismissed):
LGTM

packages/project/lib/graph/ProjectDefinitionWatcher.js
matz3: Yeah, thats right. Thanks for pointing that out. I removed it and refactored the code.

matz3:
The CI ran into a timeout on Windows, so this PR is most likely not solving all the current issues of flaky tests. I will continue investigating tomorrow.

matz3:
Moving back to draft. I will also test on Windows to verify the changes. No need to review right now.

matz3:
Flaky CI failures should be resolved via https://github.com/UI5/cli/pull/1531.

I want to follow-up on the other changes on this branch, especially suppressing the logs in our unit tests, as they can make manual checking of test failure logs harder, especially for test cases of errors.

PR #1533, opened 2026-08-21
feat: Disable generation of library manifest

d3xter666:
JIRA: CPOUI5FOUNDATION-1156

This change disables the `generateLibraryManifest` task and enforces a manifest presence for non-framework libraries

### DIFF
--- a/internal/documentation/docs/pages/Builder.md
+++ b/internal/documentation/docs/pages/Builder.md
@@ -35,7 +35,7 @@ All available standard tasks are documented under **API -> @ui5/builder -> tasks
 | [executeJsdocSdkTransformation](../api/module-@ui5_builder_tasks_jsdoc_executeJsdocSdkTransformation) |                         |                         | *disabled* <sup>1</sup> |                         |
 | [minify](../api/module-@ui5_builder_tasks_minify)                                                     |         enabled         |         enabled         |         enabled         |                         |
 | [generateFlexChangesBundle](../api/module-@ui5_builder_tasks_bundlers_generateFlexChangesBundle)      |         enabled         |         enabled         |         enabled         |                         |
-| [generateLibraryManifest](../api/module-@ui5_builder_tasks_generateLibraryManifest)                   |                         |                         |         enabled         |                         |
+| [generateLibraryManifest](../api/module-@ui5_builder_tasks_generateLibraryManifest)                   |                         |                         | *disabled* <sup>7</sup> |                         |
 | [enhanceManifest](../api/module-@ui5_builder_tasks_enhanceManifest)                                   |         enabled         |         enabled         |         enabled         |                         |
 | [generateComponentPreload](../api/module-@ui5_builder_tasks_bundlers_generateComponentPreload)        |         enabled         |         enabled         | *disabled* <sup>2</sup> |                         |
 | [generateLibraryPreload](../api/module-@ui5_builder_tasks_bundlers_generateLibraryPreload)            |                         |                         |         enabled         |                         |
@@ -58,7 +58,8 @@ All available standard tasks are documented under **API -> @ui5/builder -> tasks
 <sup>3</sup> Enabled in `self-contained` build, which disables `generateComponentPreload` and `generateLibraryPreload`  
 <sup>4</sup> Enabled for projects defining a [bundle configuration](./Configuration.md#custom-bundling)  
 <sup>5</sup> Can be enabled for framework projects via the `includeTask` option. For other projects, this task is skipped  
-<sup>6</sup> Disabled for the server due to a corresponding middleware producing the same output
+<sup>6</sup> Disabled for the server due to a corresponding middleware producing the same output  
+<sup>7</sup> Enabled for Specification Version 4.0 and lower. Not executed for Specification Version 5.0 and higher — libraries must provide a `manifest.json` in their source directory
 
 ### minify
 
--- a/internal/documentation/docs/pages/Builder.md
+++ b/internal/documentation/docs/pages/Builder.md
@@ -59,7 +59,7 @@ All available standard tasks are documented under **API -> @ui5/builder -> tasks
 <sup>4</sup> Enabled for projects defining a [bundle configuration](./Configuration.md#custom-bundling)  
 <sup>5</sup> Can be enabled for framework projects via the `includeTask` option. For other projects, this task is skipped  
 <sup>6</sup> Disabled for the server due to a corresponding middleware producing the same output  
-<sup>7</sup> Enabled for Specification Version 4.0 and lower. Not executed for Specification Version 5.0 and higher — libraries must provide a `manifest.json` in their source directory
+<sup>7</sup> Enabled for Specification Version 4.0 and lower, and for framework projects. For other projects using Specification Version 5.0 and higher, this task is skipped
 
 ### minify
 
--- a/internal/documentation/docs/pages/Configuration.md
+++ b/internal/documentation/docs/pages/Configuration.md
@@ -812,6 +812,10 @@ Version | UI5 CLI Release
 
 ### Specification Version 5.0
 
+**Breaking changes:**
+
+- The `generateLibraryManifest` build task is no longer executed by default for projects of type `library`. Libraries must provide a `manifest.json` directly in their source directory. See [Migrate to v5: generateLibraryManifest Task No Longer Executed](../updates/migrate-v5.md#generatelibrarymanifest-task-no-longer-executed) for details.
+
 **Features:**
 
 - Adds support for the new [`component`](./Project.md#component) project type for developing UI5 components — including application, reusable UI, and faceless components — which, unlike `application`-type projects, are served under their own namespace so multiple can coexist in one environment
--- a/internal/documentation/docs/pages/Configuration.md
+++ b/internal/documentation/docs/pages/Configuration.md
@@ -814,7 +814,7 @@ Version | UI5 CLI Release
 
 **Breaking changes:**
 
-- The `generateLibraryManifest` build task is no longer executed by default for projects of type `library`. Libraries must provide a `manifest.json` directly in their source directory. See [Migrate to v5: generateLibraryManifest Task No Longer Executed](../updates/migrate-v5.md#generatelibrarymanifest-task-no-longer-executed) for details.
+- The `generateLibraryManifest` build task is no longer executed for projects of type `library`. SAPUI5 distribution libraries (framework projects) are not affected. Libraries must provide a `manifest.json` directly in their source directory. See [Migrate to v5: generateLibraryManifest Task No Longer Executed](../updates/migrate-v5.md#generatelibrarymanifest-task-no-longer-executed) for details.
 
 **Features:**
 
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -46,15 +46,15 @@ Projects using older **Specification Versions** are expected to be **fully compa
 ## generateLibraryManifest Task No Longer Executed
 
 ::: info Specification Version 5.0 only
-This change only applies to library projects that upgrade their `specVersion` to `5.0` in `ui5.yaml`. Projects on **Specification Version 4.0 and lower are not affected**.
+This change only applies to library projects that upgrade their `specVersion` to `5.0` in `ui5.yaml`. Projects on **Specification Version 4.0 and lower are not affected**. **SAPUI5 distribution libraries (framework projects) are not affected**.
 :::
 
-With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed for non-framework library projects. Libraries must provide a `manifest.json` directly in their source directory.
 
 **Action required** when upgrading a library project to Specification Version 5.0:
 
 - Ensure your library has a `manifest.json` in its source directory. A previously auto-generated one is fully compatible and can be reused as-is.
-- If no `manifest.json` is present, the build will no longer generate one automatically. The build will still succeed without an error — the absence of `manifest.json` will only become apparent at runtime, where it can cause issues such as missing library dependencies, broken i18n resource loading, incorrect theme support, etc.
+- If no `manifest.json` is present, any UI5 CLI command that resolves the project graph (such as `ui5 build`, `ui5 serve`, or `ui5 tree`) will fail with a descriptive error and a link to this migration guide.
 
 ::: tip
 To see which standard tasks are executed for each project type, check out the [Standard Tasks](../pages/Builder#standard-tasks) table in the UI5 Builder page.
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -53,8 +53,8 @@ With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/modul
 
 **Action required** when upgrading a library project to Specification Version 5.0:
 
-- Ensure your library has a hand-crafted `manifest.json` in its source directory.
-- If no `manifest.json` is present, the build will no longer generate one automatically.
+- Ensure your library has a `manifest.json` in its source directory. A previously auto-generated one is fully compatible and can be reused as-is.
+- If no `manifest.json` is present, the build will no longer generate one automatically. The build will still succeed without an error — the absence of `manifest.json` will only become apparent at runtime, where it can cause issues such as missing library dependencies, broken i18n resource loading, incorrect theme support, etc.
 
 ::: tip
 To see which standard tasks are executed for each project type, check out the [Standard Tasks](../pages/Builder#standard-tasks) table in the UI5 Builder page.
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -11,8 +11,6 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 ## Breaking Changes
 
-- **@ui5/builder: `generateLibraryManifest` task is not executed for Specification Version 5.0**
-
 - **All UI5 CLI Modules require Node.js ^22.22.2 || ^24.15.0 || >=26.0.0**
 
 - **@ui5/cli: `ui5 init` defaults to Specification Version 5.0**
@@ -47,9 +45,11 @@ Projects using older **Specification Versions** are expected to be **fully compa
 
 ## generateLibraryManifest Task No Longer Executed
 
-With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
+::: info Specification Version 5.0 only
+This change only applies to library projects that upgrade their `specVersion` to `5.0` in `ui5.yaml`. Projects on **Specification Version 4.0 and lower are not affected**.
+:::
 
-For projects using **Specification Version 4.0 and lower**, the behavior is unchanged.
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
 
 **Action required** when upgrading a library project to Specification Version 5.0:
 
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -11,6 +11,8 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 ## Breaking Changes
 
+- **@ui5/builder: `generateLibraryManifest` task is not executed for Specification Version 5.0**
+
 - **All UI5 CLI Modules require Node.js ^22.22.2 || ^24.15.0 || >=26.0.0**
 
 - **@ui5/cli: `ui5 init` defaults to Specification Version 5.0**
@@ -42,6 +44,21 @@ Support for older Node.js releases has been dropped; their use will cause an err
 UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Component Type and brings improved project structure conventions.
 
 Projects using older **Specification Versions** are expected to be **fully compatible with UI5 CLI v5**.
+
+## generateLibraryManifest Task No Longer Executed
+
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
+
+For projects using **Specification Version 4.0 and lower**, the behavior is unchanged.
+
+**Action required** when upgrading a library project to Specification Version 5.0:
+
+- Ensure your library has a hand-crafted `manifest.json` in its source directory.
+- If no `manifest.json` is present, the build will no longer generate one automatically.
+
+::: tip
+To see which standard tasks are executed for each project type, check out the [Standard Tasks](../pages/Builder#standard-tasks) table in the UI5 Builder page.
+:::
 
 ## Build Cache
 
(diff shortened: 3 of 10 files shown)


d3xter666:
Hi @KlattG 

Would you take a look at the modified docs:
- internal/documentation/docs/pages/Builder.md ([Rendered version](https://github.com/UI5/cli/blob/f00c99b2d01e8d0f22a23342cd712225ab6bbb30/internal/documentation/docs/pages/Builder.md))
- internal/documentation/docs/updates/migrate-v5.md ([Rendered version](https://github.com/UI5/cli/blob/f00c99b2d01e8d0f22a23342cd712225ab6bbb30/internal/documentation/docs/updates/migrate-v5.md))

Cheers

packages/builder/lib/tasks/generateLibraryManifest.js
```diff
@@ -11,6 +11,9 @@ import manifestCreator from "../processors/manifestCreator.js";
 /**
  * Task for creating a library manifest.json from its .library file.
  *
+ * **Note:** This task is not executed for projects using Specification Version 5.0 or higher.
```
matz3: The task itself is independent from the decision not to run it in certain scenarios. If you call it directly, you can execute it for a library project with specVersion 5.0 or higher. Same as we will still call it for the framework libraries.

As the task itself stays independent, I would not mention anything in this JSDoc here.

internal/documentation/docs/pages/Builder.md
```diff
@@ -58,7 +58,8 @@ All available standard tasks are documented under **API -> @ui5/builder -> tasks
 <sup>3</sup> Enabled in `self-contained` build, which disables `generateComponentPreload` and `generateLibraryPreload`  
 <sup>4</sup> Enabled for projects defining a [bundle configuration](./Configuration.md#custom-bundling)  
 <sup>5</sup> Can be enabled for framework projects via the `includeTask` option. For other projects, this task is skipped  
-<sup>6</sup> Disabled for the server due to a corresponding middleware producing the same output
+<sup>6</sup> Disabled for the server due to a corresponding middleware producing the same output  
+<sup>7</sup> Enabled for Specification Version 4.0 and lower. Not executed for Specification Version 5.0 and higher — libraries must provide a `manifest.json` in their source directory
```
matz3: This note is missing to explain the exception for framework projects (similar to `<sup>5</sup>`)

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -11,6 +11,8 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 ## Breaking Changes
 
+- **@ui5/builder: `generateLibraryManifest` task is not executed for Specification Version 5.0**
```
matz3: I'm not sure whether this change is a breaking change of `@ui5/builder`, especially as the PR only adds JSDoc comments to files in the `builder` package, but does not adjust any code.

In addition, I'm also not sure whether this PR is a breaking change in general. It is a breaking change that comes with specVersion 5.0, but not a change that needs adjustment by consumers, unless they upgrade to specVersion 5.0. This is an important difference which allows us to make sure changes without disrupt existing projects.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +45,21 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
 
 Projects using older **Specification Versions** are expected to be **fully compatible with UI5 CLI v5**.
 
+## generateLibraryManifest Task No Longer Executed
+
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
```
matz3: Should we also mention the breaking change in the Specification Version documentation at `internal/documentation/docs/pages/Configuration.md` ?

packages/builder/lib/tasks/generateLibraryManifest.js
d3xter666: Removed the comment entirely

internal/documentation/docs/updates/migrate-v5.md
```suggestion
- **@ui5/builder: The `generateLibraryManifest` task is not executed for Specification Version 5.0**
```

KlattG (dismissed):
One minor suggestion, otherwise fine

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I have now removed the breaking change information from here and moved it to the specVersion docs.

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Yep, I have added a section in the `specVersion` section in the Configuration

internal/documentation/docs/pages/Builder.md
d3xter666: I have adjusted it now

internal/documentation/docs/updates/migrate-v5.md
d3xter666: This one has been entirely removed, since we do not consider this a breaking change in the existing functionality

d3xter666:
Not quite sure whether we need to keep the `BREAKING CHANGE` in the description. It's breaking as of `specVersion: 5` and is mentioned in the migration guide. But as we discussed, existing projects with specVersions < 5 are not affected

internal/documentation/docs/pages/Configuration.md
```diff
@@ -812,6 +812,10 @@ Version | UI5 CLI Release
 
 ### Specification Version 5.0
 
+**Breaking changes:**
+
+- The `generateLibraryManifest` build task is no longer executed by default for projects of type `library`. Libraries must provide a `manifest.json` directly in their source directory. See [Migrate to v5: generateLibraryManifest Task No Longer Executed](../updates/migrate-v5.md#generatelibrarymanifest-task-no-longer-executed) for details.
```
matz3: `no longer executed by default` can give the impression that there is a way to still execute it. But I assume there isn't, right? (except for the framework projects, which are not mentioned here but maybe should be mentioned?).

Background:
Using `--include-task generateLibraryManifest` won't make a difference as this change makes the task a "no-op" instead of disabling it (composeTaskList). The task still exists and is included by default (so custom tasks can still refer to it, and that it still runs for specVersion < 5.0 / framework projects in the tree), but otherwise it is a no-op.
I think this is the best solution without inventing new concepts or changing a lot of existing code.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
...
+
+::: info Specification Version 5.0 only
+This change only applies to library projects that upgrade their `specVersion` to `5.0` in `ui5.yaml`. Projects on **Specification Version 4.0 and lower are not affected**.
+:::
+
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed. Libraries must provide a `manifest.json` directly in their source directory.
+
+**Action required** when upgrading a library project to Specification Version 5.0:
+
+- Ensure your library has a hand-crafted `manifest.json` in its source directory.
```
matz3: Wouldn't it be a good suggestion to run a build and copy the generated `manifest.json` from the build output before changing the specVersion to 5.0? I don't see a reason why people need to hand-craft it. They just need to check it into their sources now and make changes manually, but using the generated one would restore the previous behavior without much effort.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
...
+- If no `manifest.json` is present, the build will no longer generate one automatically.
```
matz3: Do we know which consequences this would have? I tested it and did not get any warning or error, so I would assume people can easily miss this when upgrading to specVersion 5.0 without checking all details.

packages/project/lib/build/definitions/library.js
```diff
@@ -95,7 +95,12 @@ export default function({project, taskUtil, getTask}) {
 		}
 	});
 
-	tasks.set("generateLibraryManifest", {});
+	// For specVersion 5.0+, only execute for framework libraries
+	if (project.getSpecVersion().lt("5.0") || project.isFrameworkProject()) {
+		tasks.set("generateLibraryManifest", {});
+	} else {
+		tasks.set("generateLibraryManifest", {taskFunction: null});
```
matz3, line 102: Apparently, the cache is not invalidated if a project now runs into this line. This means the old cache will be re-used (e.g. when already using specVersion 5.0 and then upgrading to a new pre-release that contains this change).

Those are the changes we have to watch out for, as using a stale cache is a bug that is often hard to find. We should maybe even invest into automated tests to ensure caching works as expected.

We have different options to resolve this. Bumping `BUILD_SIG_VERSION` would be the easiest (but invalidates all existing caches). Adding a `determineBuildSignature` callback would be a more fine-grained solution, which I would rather prefer. Otherwise we could think of generically detecting this by adding the "taskFunction" type (function / null) into the default signature.

matz3:
> Not quite sure whether we need to keep the `BREAKING CHANGE` in the description. It's breaking as of `specVersion: 5` and is mentioned in the migration guide. But as we discussed, existing projects with specVersions < 5 are not affected

Yes, `BREAKING CHANGE` should be removed, because upgrading the specVersion is not bound to upgrading the npm package. We should only describe breaking changes in commits (conventional commits) if something changes when upgrading the npm package. If we would merge this PR when we already have a stable v5 released, we would also not bump the version to v6. Therefore, not a breaking change.

packages/project/lib/build/definitions/library.js
d3xter666: Hmmm, thanks!
This I didn't consider!
However, it seems like something we need to resolve in general. The case with `generateBundle`, `generateComponentPreload`, etc... (tasks being registered with different arguments or noop callbacks in general).

From that perspective, although it's relevant, my feeling is that, it can grow eventually in complexity and I would rather prefer to resolve this separately, so we can discuss there potential drawbacks and architectural discussions.

internal/documentation/docs/pages/Configuration.md
d3xter666: Thanks! I have removed the `no longer executed by default` and have added the framework libs.
I think this brief message should be enough here since we have a link to the v5 migration guide where more details and context should be provided, so that developers properly migrate their code

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I agree that the consequences of an absent `manifest.json` are not sufficiently documented here, and I will add information about what this can cause at runtime.

Regarding the warning, the same pattern currently applies to any task that is skipped via `tasks.set("taskName", { taskFunction: null });`.

At the moment, tasks skipped this way provide a warning only in --verbose mode:
https://github.com/UI5/cli/blob/62a3e3971ca3079b7cbfb6cfd4c28bd2e902e767/packages/project/lib/build/TaskRunner.js#L199

I suggest we keep this behavior consistent with the existing pattern rather than extending the scope of this PR. 

Edit: IMO, simply a warning would not be enough as, in this case we must first check for an existing `manifest.json` and then warn.

internal/documentation/docs/updates/migrate-v5.md
d3xter666: Thanks! Updated the message

packages/project/lib/build/definitions/library.js
matz3: Yes, this applies to all tasks here, and we need to check that for all changes we make (also within the code of the tasks/processors). However, I do not see current issues with `generateBundle` / `generateComponentPreload` unless we make new changes. All data they rely on is already part of the signature, so it will be invalidated properly if something changes.

The problem here is rather how to make sure that code changes are reflected in the signature, and I agree that this should generally be discussed and defined.

And as long as we are in pre-release state, I assume such issues aren't as severe as they would be once the release is stable.

internal/documentation/docs/updates/migrate-v5.md
matz3: My concern was not about seeing that tasks are skipped, but rather that we should work towards always having a `manifest.json` available for libraries at runtime. Otherwise I see this change as one step back instead of forward. Yes, if you skipped the task explicitly you would already have the same state, but skipping it is nothing we recommend so it would be a user error.

This should be seen as part of the specVersion change of this BLI and we should now think about and define what consequences our change has or should have.
For type `application` we already simply fail with an error if there is no `manifest.json`, which means we can be sure that there will always be a `manifest.json`.
If we now allow projects with specVersion 5.0 to not have a `manifest.json`, we cannot rely on having one available. And if we do not do it now (i.e. ideally before the next pre-release; definitely before the stable release), failing the build / making it mandatory would be another specVersion change or breaking CLI change.

I hope you get my point that I think those decisions should be taken now. This is something we did not discuss so far, because at least for me it only became clear now, when I was reviewing the code and thinking about consequences of this change.
And of course it must not be all in one PR, but it is something that should not be neglected or ignored.

d3xter666: Thanks!
I didn't know about this handling for the Application!

I have now aligned them with https://github.com/UI5/cli/pull/1533/commits/7cc5d60f6d8191d727dc2e95af109218564b7201

packages/project/lib/build/definitions/library.js
RandomByte: I created https://github.com/UI5/cli/pull/1550 which incorporates the `@ui5/project` package version into the build signature (and also fixes the usage of our other package versions). This should resolve that issue.

packages/project/lib/specifications/types/Library.js
```diff
@@ -194,6 +194,23 @@ class Library extends ComponentProject {
 	async _parseConfiguration(config, buildManifest) {
 		await super._parseConfiguration(config, buildManifest);
 
+		// For Specification Version 5.0+, non-framework libraries must provide a manifest.json
+		if (this.getSpecVersion().gte("5.0") && !this.isFrameworkProject()) {
+			try {
+				await this._getManifest();
```
matz3: Here the check is done before the `if (buildManifest) {` check and early return, but for `Application.js` / `Component.js` the check / error handling is done after (via `_getNamespace`). I'm not sure whether it makes a difference, but if it does, I assume we should align both places, right?

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
 
 Projects using older **Specification Versions** are expected to be **fully compatible with UI5 CLI v5**.
 
+## generateLibraryManifest Task No Longer Executed
```
matz3, line 46: This section should be reviewed again by UA, but my thoughts would be:
Does the title still explain the changes well enough? It is not wrong, but something that describes the new expectations with specVersion 5, i.e. "manifest.json is mandatory" might be easer to understand for consumers, who might not even have heard of the task name before.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
...
+
+::: info Specification Version 5.0 only
+This change only applies to library projects that upgrade their `specVersion` to `5.0` in `ui5.yaml`. Projects on **Specification Version 4.0 and lower are not affected**. **SAPUI5 distribution libraries (framework projects) are not affected**.
+:::
+
+With **Specification Version 5.0**, the [`generateLibraryManifest`](../api/module-@ui5_builder_tasks_generateLibraryManifest) build task is no longer executed for non-framework library projects. Libraries must provide a `manifest.json` directly in their source directory.
```
matz3, line 52: If you take `directly in their source directory` literally, then it would be at `src/manifest.json`, right? This would be wrong if the library uses folders for the namespace (like `src/sap/ui/core/`). I think we should make sure the expectations are described in a way that they cannot be misunderstood.
Saying that it needs to be placed next to `library.js` would be a good way of explaining what to do, without having to explain the two different ways of structuring the sources.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
...
+
+**Action required** when upgrading a library project to Specification Version 5.0:
+
+- Ensure your library has a `manifest.json` in its source directory. A previously auto-generated one is fully compatible and can be reused as-is.
```
matz3, line 56: I would rather have described this more literally, by telling to run a build before changing the specVersion and copy over the `manifest.json` file (located next to the `library.js` file) from the build result into the sources.

packages/project/lib/specifications/types/Library.js
d3xter666: `buildManifest` is not the manifest itself in some form. Please, take a look at the JSDoc: `@param {object} buildManifest Cache metadata object`.

There's a manifest check in the `_getNamespace()` -> `_getNamespaceFromManifest()`, but it can easily be skipped i.e. this early return for `buildManifest`

Yes, `_getNamespace()` requests the manifest, but its sole purpose it to extract the namespace. Therefore, messing in there would make the solution quite more complex. 
For Application type we do not have these constraints we have for Libraries!

Given that, my preference would be to keep this solution as clearer and more maintainable

internal/documentation/docs/updates/migrate-v5.md
d3xter666: From my perspective, it states clear what we have done and what action needs to be taken- we have disabled *{some task name}* and `manifest.json` is now mandatory when building a library. Further more, https://github.com/UI5/cli/pull/1533/commits/7cc5d60f6d8191d727dc2e95af109218564b7201 enforces develoeprs to provide a `manifest.json`

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I don't want to judge on developers' experience and knowledge. It's clear now that they need to provide a `manifest.json` for the library project. It's event enforced now with https://github.com/UI5/cli/pull/1533/commits/7cc5d60f6d8191d727dc2e95af109218564b7201.

IMO, we must not mix/couple our documentation with UI5 runtime information. It must be clear from the UI5 documentation where certain files must be positioned and all the edge cases that might come out of that.

internal/documentation/docs/updates/migrate-v5.md
d3xter666: I don't feel this is the correct way to describe this. For me, it feels like a workaround over the old `specVersion`, so that you can run the code with the new one.
Then, if something had a better experience previously, why do we do this migration?

packages/project/lib/specifications/types/Library.js
matz3: That is not what I meant. Yes, `buildManifest != manifest.json`.

If an application project is already built, and therefore `buildManifest` is supplied, then we do not perform a check for its `manifest.json` file because we rely on the build that happened before to check for such constraints. The buildManifest provides the namespace, so we can use it from there.
The same happens for component projects.

But now if a library project is already built, the existence of a `manifest.json` is still checked.
This is my point. I have not checked whether it causes issues (or if it is even covered by tests), but to me this is not consistent with our existing handling in other types.

d3xter666: Yes, and now, after this task, the `manifest.json` is not part of the build anymore. It is additionally provided as a resource.
It must **not** be treated as a build artifact or something. As long as we rely on this fact, we must check its existence everytime.
As far as I got it from the code, the namespace for `buildManifest` might **not** come from the `manifest.json`

internal/documentation/docs/updates/migrate-v5.md
flovogt, line 56:
```suggestion

- Ensure your library has a `manifest.json` in its source directory. If you would like to leverage the power of UI5 CLI, you could execute `ui5 build` using Specification Version `4.0` and copy the generated `manifest.json` file from the build result to the source directory of your library.
```

internal/documentation/docs/updates/migrate-v5.md
flovogt: Agreed with UA @KlattG. Go for the current used variant.

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -43,6 +43,23 @@ UI5 CLI 5.x introduces **Specification Version 5.0**, which enables the new Comp
...
+- If no `manifest.json` is present, any UI5 CLI command that resolves the project graph (such as `ui5 build`, `ui5 serve`, or `ui5 tree`) will fail with a descriptive error and a link to this migration guide.
```
flovogt, line 57:
```suggestion
- If no `manifest.json` is present, UI5 CLI commands such as `ui5 build`, `ui5 serve`, or `ui5 tree` will fail with an error.
```

flovogt (commented):
UA review done by @KlattG and  @flovogt

packages/project/lib/specifications/types/Library.js
RandomByte: I guess the current implementation is fine. But I agree that the asymmetry with what is essentially the same check (manifest.json must exist) in the Application and Component types is confusing. Also the error message differs.

As Matthias already mentioned, if a build manifest exists, the project has already been built by UI5 CLI and is therefore _guaranteed_ to contain a `manifest.json`. Of course we can check again, but then why don't we do that for e.g. the Component type?

Side remark: the Application and Component types currently contain copies of the same `_getNamespaceFromManifestJson` method. Someone keen on optimizing our code could check why that is and then think about moving it into the `ComponentProject` superclass to create reuse across all three types now.

RandomByte (dismissed):
Implementation is acceptable. I didn't review the doc changes

d3xter666:
Thank you for your review, @RandomByte!

I just wanted to clarify certain things as I still have some concerns regarding this:
> As Matthias already mentioned, if a build manifest exists, the project has already been built by UI5 CLI and is therefore guaranteed to contain a manifest.json. Of course we can check again, but then why don't we do that for e.g. the Component type?

Here's my interpretation and why I have build it that way:
The `buildManifest` stores the computed namespace, not a guarantee that the source `manifest.json` still exists. If someone deletes `manifest.json` after a successful build that wrote a `build-manifest.json`, the `buildManifest` would be stale. That's why, IMO we need this explicit check.

I also saw that the `Application` and `Component` try out the `manifest.appdescr_variant`! In the BLI I didn't see something about it. Is it applicable also here?

packages/project/lib/specifications/types/Library.js
d3xter666: After the sync, the outcome is that as long as there is a `buildManifest`, developers must ensure existence of the `manifest.json` since it's been there during the `buildManifest` creation. And this part was missing to me.

I have revised the implementation and now it's close to the one in `Component` and `Application`

RandomByte (approved):
Implementation LGTM. Did not review the docs as agreed.

packages/project/lib/specifications/types/Library.js
```diff
@@ -391,9 +391,22 @@ class Library extends ComponentProject {
...
-				`${this.getName()}: ${err.message}`);
+			// For Specification Version 5.0+, non-framework libraries must provide a manifest.json
+			if (this.getSpecVersion().gte("5.0") && !this.isFrameworkProject()) {
+				throw new Error(
+					`Could not find required manifest.json for library project ${this.getName()}: ` +
+					`${err.message}\n\n` +
+					`Library projects using Specification Version 5.0 or higher must provide ` +
+					`a manifest.json directly in their source directory.\n` +
+					`For migration details, please refer to:\n` +
+					`https://ui5.github.io/cli/updates/migrate-v5/#generatelibrarymanifest-task-no-longer-executed`, {
```
matz3, line 402: For me, even after merging the PR, this redirects to https://ui5.github.io/cli/stable/404.html#generatelibrarymanifest-task-no-longer-executed

d3xter666: Yep, migrate-v5 does not exist in v4 documentation.
Thanks, I will update it to point to the next doc branch!

d3xter666: Fixed with https://github.com/UI5/cli/pull/1553

PR #1540, opened 2026-08-24
ci: Prevent internal packages from declaring production dependencies

matz3:
Add a CI check that fails if any "internal/*" workspace package declares a production "dependencies" entry. All dependencies of these packages must be declared as "devDependencies".

Dependabot derives the commit-message type from whether an update targets "dependencies" or "devDependencies" ("deps" vs. "build(deps-dev)") and does not treat our internal, unpublished packages differently. A production dependency there therefore produces a misleading "deps" commit.

Migrate the existing production dependencies of the benchmark, documentation, e2e-tests and shrinkwrap-extractor internal packages to devDependencies, which is safe as these packages are private and never published.

Consolidate this check and the existing "engines" consistency check into a single reusable scripts/check-workspaces.js (run via "npm run check-workspaces"), replacing the two inline Node heredocs in github-ci.yml with one step.

Document the rule for contributors and agents in AGENTS.md.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -33,27 +33,8 @@ jobs:
     - name: Check Node.js engine compatibility
       run: npm run check-engine
 
-    - name: Check package.json "engines" consistency
-      run: |
-        node << 'EOF'
-        const { execSync } = require('child_process');
-        const fs = require('fs');
-
-        const rootPkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
-        const workspaces = JSON.parse(execSync('npm query .workspace --json', { encoding: 'utf8' }));
-
-        let hasError = false;
-        workspaces.forEach(pkg => {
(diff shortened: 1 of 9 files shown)


.github/workflows/github-ci.yml
```diff
@@ -55,6 +55,34 @@ jobs:
...
+            console.error(`❌ ${pkg.location}/package.json declares production "dependencies" ` +
+              `(${prodDeps.join(', ')}). Internal packages must declare all dependencies as "devDependencies".`);
+            hasError = true;
+          }
+        });
+
+        if (hasError) process.exit(1);
+        console.log('✅ No internal package declares production dependencies');
+
+        EOF
```
d3xter666: Can we consolidate this and the previous script (`Check package.json "engines" consistency`) into some reusable script like this one `scripts/unit-summary.js`

They seem pretty similar, except for checking different fields

matz3: Yeah you are right, thanks. Having a script is much better, as you can now also run it locally.

d3xter666 (approved):

PR #1541, opened 2026-08-24
ci: Group Dependabot updates and adjust cooldown

matz3:
- Group all github-actions updates into one PR
- Group npm minor/patch updates into one PR
- Bump cooldown to 3 days for both ecosystems
- Exclude @ui5/* and less-openui5 from npm cooldown

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,10 @@ updates:
   commit-message:
     prefix: "ci(github-actions)"
   cooldown:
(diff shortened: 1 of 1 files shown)


d3xter666 (approved):

matz3:
It seems like groups have limitations like not being able to configure the commit message prefixes. Let's still try this out, but I fear we have to revert the groups again if we want conventional commit messages.

PR #1551, opened 2026-08-27
refactor: Convert shrinkwrap extractor tool to package-lock

d3xter666:
JIRA: CPOUI5FOUNDATION-1283

Shrinkwrap is being deprecated (https://github.com/npm/cli/pull/9262) in node, but we want to keep and guard the dependencies that we ship with the `UI5 CLI`.

We have decided to go with NPM's packge.json [`bundeDependencies` property](https://docs.npmjs.com/cli/v12/configuring-npm/package-json#bundledependencies)

`bundleDependencies` has its own issues when working with monorepos. The core issue is that it simply bundles anything that's within that package's `node_modules/` folder and does not consider any hoisted packages that monorepos naturally have.

With this context in mind, we have decided to leverage our internal tool `shrinkwrap-extractor` and adjust it to fit for the `bundleDependencies` case.

Here's the expected workflow:
- Rename `shrinkwrap-extractor` to something more meaningful in the new context. **Proposal**: `lockfile-extractor`
- Keep the existing CI publish flow (https://github.com/UI5/cli/blob/main/.github/workflows/release-please.yml#L134-L144) and run  `lockfile-extractor` for `packages/cli`. This will result in `package-lock.json` file extraction suitable just for the `@ui5/cli` package.
- Move `packages/cli` out of the scope of the monorepo by copying its files in a $TEMP dir.
- Within the $TEMP dir, remove any non relevant dependencies (`devDependencies`) from `package.json`
- Run `npm ci`, so that dependencies are installed with the exact same versions that are defined within the generated earlier `package-lock.json`
- Run `npm pack` and `npm publish`, so that the generated tarball taht includes also the `bundledependencies` gets published on NPM

---

**Note:** This PR simply renames the _**shrinkwrap-extractor**_ to _**lockfile-extractor**_. The only real change is the handling of publish GH Action: https://github.com/UI5/cli/pull/1551/changes#diff-2c84033033d49186c63e6adcd705f63b11ae6814cd76c152c9c486d389fbccf3


internal/lockfile-extractor/lib/extractFromWorkspaceLockfile.js
```diff
@@ -0,0 +1,340 @@
+import {readFile, mkdtemp, writeFile, rm} from "node:fs/promises";
```
d3xter666: This file is actually renamed `internal/shrinkwrap-extractor/lib/convertPackageLockToShrinkwrap.js`. I don't know why GitHub splits them this way

.github/workflows/release-please.yml
```diff
@@ -131,14 +131,34 @@ jobs:
...
+
+          cd "$TEMP_CLI_DIR"
+
+          # Strip devDependencies so npm ci only installs production deps matching the lock file
+          node -e "
+          const fs = require('fs');
+          const pkg = JSON.parse(fs.readFileSync('package.json', 'utf8'));
+          delete pkg.devDependencies;
+          fs.writeFileSync('package.json', JSON.stringify(pkg, null, '\t'));
+          "
```
d3xter666: We need this explicit `devDependencies` strip since the `lockfile-extractor` provides only the prod dependencies. Any `npm ci` (even `npm ci --omit=dev`) call wil fail if we don't do this.

.github/workflows/bundle-and-publish-cli.yml
```diff
@@ -0,0 +1,85 @@
+name: Bundle and Publish @ui5/cli
```
d3xter666, line 1: This is an actual copy of `shrinkwrap/lockfile-extractor` run + package deployment to NPM from `release-please.yaml`

Now it has a dry run flag, so we can integrate this flow into our current CI environment. This way, we can ensure that every merge in the `main` branch won't break the release process.

.github/workflows/bundle-and-publish-cli.yml
```diff
@@ -0,0 +1,85 @@
...
+
+jobs:
+  bundle-and-publish:
+    runs-on: ubuntu-24.04
+    # Controls the OIDC subject claim: 'repo:UI5/cli:environment:<npm_env>'.
+    # npm's trusted publisher verifies this claim — empty string means no environment
+    # (subject becomes 'repo:UI5/cli:ref:refs/heads/...'), used for dry-run only.
+    environment: ${{ inputs.npm_env }}
+    permissions:
+      id-token: write # Required for OIDC trusted publishing; harmless when npm_env is empty
```
RandomByte: I was wondering whether this would allow PRs to manipulate the workflow file and execute a (non-dry-run) publish, but it seems like this is not possible since we protect the `npmjs:ui5-cli-mono` environment 👍 

But this also means that the id-token permission declared here and in the `github-ci.yml` might not be needed, since the token can't be used to authenticate with the registry anyways (see the log message `This command requires you to be logged in to https://registry.npmjs.org/ (dry-run)` in the test run https://github.com/UI5/cli/actions/runs/33755786322/job/100652793446#step:6:443 which comes from here: https://github.com/npm/cli/blob/b016aa287d01fcd688d6704222f193ce5c04bfb7/lib/commands/publish.js#L167

According to the GitHub docs, permissions are inherited to called jobs, so we could do this cleanup. But definitely already safe the way it is right now.

.github/workflows/bundle-and-publish-cli.yml
```diff
@@ -0,0 +1,85 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
RandomByte, line 41:
```suggestion
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
        with:
          persist-credentials: false
```

Safeguard since we run a lot of scripts here and there is no need to push anything anyways.

.github/workflows/bundle-and-publish-cli.yml
```diff
@@ -0,0 +1,85 @@
...
+          fs.writeFileSync('package.json', JSON.stringify(pkg, null, '\t'));
+          "
+
+          # Install exact versions from generated lock file (no workspace symlinks)
+          echo "📦 Installing production dependencies from package-lock.json"
+          npm ci
+
+          if [ "${{ inputs.dry_run }}" = "true" ]; then
+            echo "🔍 Dry-run: reporting what would be published (no actual publish)"
+            npm publish --access public --tag "${{ inputs.npm_tag }}" --dry-run
```
RandomByte: [zizmor](https://github.com/zizmorcore/zizmor) (one of the OSPO tools for scanning our actions) reports this as a potential [template injection](https://docs.zizmor.sh/audits/#template-injection). It's actually safe for us due to how we invoke this workflow, but I think we should rather fix the finding.

A [possible fix](https://securitylab.github.com/resources/github-actions-untrusted-input/#remediation) would be to add an `env:` property above, declaring env-vars for the inputs like this and use those in the command executions:

```yml
      - name: Bundle and publish @ui5/cli
        env:
          NPM_TAG: ${{ inputs.npm_tag }}
```

Command:
`npm publish --access public --tag "$NPM_TAG" --dry-run`

.github/workflows/github-ci.yml
```diff
@@ -86,6 +86,15 @@ jobs:
         npm run build:vitepress
         npm run build:assets
 
-    - name: Check shrinkwrap integrity
-      working-directory: internal/shrinkwrap-extractor
+    - name: Check lockfile extractor integrity
```
RandomByte: Should this be something like the below? Or what integrity are we checking here?

```suggestion
    - name: Run lockfile extractor tests
```

RandomByte (changes requested):
I'm still think we might not handle all cases of dependency hoisting correctly, as the added test in https://github.com/UI5/cli/pull/1569 was supposed to show. At the same time, I don't see how this process would catch such problems. Therefore I'm hesitant to approve this without any planned follow-up work to address those concerns.

`npm ci` will install packages as specified in the generated `package-lock.json`. I believe it _does not validate_ whether hoisted dependencies match with the nested project's requirements (i.e. whether the hoisted dependency's version matches the specified range).

For `npm publish` I believe that it would pack whatever is provided in `node_modules`, again without further validation whether that matches with the dependency's requirements.

I might be wrong on both points here, and I would appreciate to get some well founded reassurance if that is the case. Otherwise I fear that the package-lock.json we "extract" would silently lead to incorrect dependency versions getting installed. Note that this can not be tested at runtime. Only by validating the dependency tree.

.github/workflows/bundle-and-publish-cli.yml
d3xter666: Thanks! 
I have fixed it with https://github.com/UI5/cli/pull/1551/changes/6d4ad5147b0abe9e446ef6752aea6df72bf6e009

.github/workflows/bundle-and-publish-cli.yml
d3xter666: Thanks!
Refactored permissions with https://github.com/UI5/cli/pull/1551/commits/a4874fb38c9aae9eea9556295afcd4a662dae849

internal/lockfile-extractor/package.json
```diff
@@ -0,0 +1,48 @@
...
+		"directory": "internal/lockfile-extractor"
+	},
+	"scripts": {
+		"test": "npm run lint && npm run coverage && npm run integration && npm run depcheck",
+		"unit": "node --test test/lib/extractFromWorkspaceLockfile.js",
+		"unit-watch": "node --test --watch test/lib/extractFromWorkspaceLockfile.js",
+		"integration": "node --test test/integration/lockfile-validation.js",
+		"coverage": "mkdir -p coverage && node --test --experimental-test-coverage --test-reporter=spec --test-reporter-destination=stdout --test-reporter=lcov --test-reporter-destination=coverage/lcov.info test/lib/extractFromWorkspaceLockfile.js",
+		"lint": "eslint .",
+		"depcheck": "depcheck --ignores @ui5/lockfile-extractor"
```
matz3: Why do we suddenly have depcheck in here? Didn't we remove that in favor of knip? This make me wonder what revision of "shrinkwrap-extractor" was used here, and might even explain why git fails to detect the renames, as the content differs too much.

internal/lockfile-extractor/LICENSES/Apache-2.0.txt
matz3, line 1: It is just a rename (pre-existing finding), but I do not think we have to keep the license in here. We already have the file present in our repo.

d3xter666: Done with https://github.com/UI5/cli/pull/1551/commits/57b8fe16c4455f84192357c1119e527cc27d17dc

internal/lockfile-extractor/package.json
d3xter666: Thanks! Now it should be a lot better. I think I managed to rename them properly this time

internal/lockfile-extractor/LICENSES/Apache-2.0.txt
d3xter666: Sorry, I need to revert it. It seems that we check each package individually: https://github.com/UI5/cli/blob/main/.github/workflows/reuse-compliance.yml#L19

https://github.com/UI5/cli/pull/1551/commits/d480b7d0d5bbce760d13b8942664bb6f8fc2534e

matz3: Oh sorry. Fine we me to keep it then.
I looked at benchmark and e2e-tests, which both are not included in that reuse check. Not sure why we have not added them, but that's nothing we have to address here.

package-lock.json
```diff
@@ -93,16 +93,17 @@
...
 			"version": "1.0.0",
 			"license": "Apache-2.0",
 			"bin": {
-				"shrinkwrap-extractor": "cli.js"
+				"lockfile-extractor": "cli.js"
 			},
 			"devDependencies": {
 				"@npmcli/arborist": "^10.0.2",
 				"@npmcli/config": "^11.0.1",
+				"depcheck": "^1.4.7",
```
matz3: Lockfile still looks out of sync

d3xter666: Thanks!
I have updated the lockfile.
Now should be ok

d3xter666:
Closing this PR. Rationale: https://github.com/UI5/cli/discussions/1408

PR #1552, opened 2026-08-27
refactor(project): Remove unused ProjectBuildCache.create factory

RandomByte:
This is a leftover from an earlier refactoring.

`ProjectBuildContext.create()` constructs the cache with `new ProjectBuildCache(...)` directly (ProjectBuildContext.js:85). After the async initialization moved out into `initSourceIndex()`, the static `create()` factory was reduced to `return new ProjectBuildCache(...)` and its only remaining callers were the unit tests.

* Remove the factory and have the tests construct via the constructor.
* Update the constructor and initSourceIndex doc comments that still referenced `create()`.


d3xter666 (approved):
LGTM

PR #1554, opened 2026-08-28
ci: Add npm script for docs in root

maxreichmann:
Adds a simple dev script **in root** for **generating the docs and starting the dev server**.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,6 @@
 		"knip": "knip --config knip.config.js",
 		"check-engine": "check-engine-light .",
 		"check-licenses": "licensee --errors-only",
-		"schema-generate": "npm run schema-generate --workspace=@ui5/documentation",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
 		"check-licenses": "licensee --errors-only",
 		"schema-generate": "npm run schema-generate --workspace=@ui5/documentation",
 		"generate-cli-doc": "npm run generate-cli-doc --workspace=@ui5/documentation",
```
matz3: For me it would be fine to remove those two from the root package. I assume that the common use case is `npm run docs:dev`, and for anything else, you can run the scripts in documentation.
Or if we want a script here, I would propose one script to build all documentation without starting the server. This would also be essential for agents that want to build and check the output without running a server.

What do you think?

matz3 (approved):

PR #1555, opened 2026-08-28
refactor(cli): Consolidate UI5 data dir resolution into dataDir module

matz3:
Introduce lib/dataDir.js exposing getUi5DataDir, getUi5DataDirOrDefault and a formatPath helper that shortens home-directory paths to ~ for display. This consolidates the previously duplicated ui5DataDir resolution logic that framework utils and the cache command each carried their own copy of.

The cache command now resolves the data dir via getUi5DataDirOrDefault and lazily imports FrameworkCache and CacheManager, so the modules are not loaded on every CLI invocation. Displayed cache paths use formatPath.

This only consolidates resolution within the cli package; a general, cross-package cleanup remains tracked in PR #1456.

### DIFF
--- a/packages/cli/lib/cli/commands/cache.js
+++ b/packages/cli/lib/cli/commands/cache.js
@@ -1,12 +1,9 @@
 import chalk from "chalk";
 import path from "node:path";
-import os from "node:os";
 import process from "node:process";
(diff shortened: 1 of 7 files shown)


d3xter666 (approved):
LGTM

PR #1532, opened 2026-08-21
feat!: Replace HTTP/2 support with HTTPS

matz3:
Remove the `--h2` option of `ui5 serve` and the underlying HTTP/2 infrastructure, and add a new `--https` option in its place.

HTTP/2 was served via the `spdy` package, which is unmaintained and does not work with Node.js v24 and higher. The new `--https` option uses Node.js' built-in `https` server instead. The existing SSL certificate handling (auto-generation, `--key`/`--cert`) is reused.

The HTTPS test disabled certificate validation via NODE_TLS_REJECT_UNAUTHORIZED=0.
Replace this with proper trust of the server certificate via superagent's .ca() on the test agent, applied once so all requests inherit it.
The previous fixture server.crt was a leaf certificate signed by a separate, uncommitted CA, so it could not act as its own trust anchor. Regenerate it as a genuine self-signed certificate for localhost.

JIRA: CPOUI5FOUNDATION-1310
JIRA: CPOUI5FOUNDATION-1311
See: https://github.com/UI5/cli/issues/327
Closes: https://github.com/UI5/cli/pull/1131

BREAKING CHANGE: The `--h2` option of `ui5 serve` has been removed. Use `--https` to serve the project over HTTPS. The `@ui5/server` `serve()` option `h2` and the returned `h2` property have been renamed to `https`. If you need HTTP/2 for local development, put a reverse proxy (e.g. nginx) in front of the server to terminate it.

PR #1560, opened 2026-08-31
feat(cli)!: Add "certificate generate" command

matz3:
Add "ui5 certificate generate" to create and install the self-signed server certificate before running the dev server. The command honors UI5_DATA_DIR and the configured UI5 data dir, skips existing certificates unless --force is used, and prints the written paths.

Change the common name of the certificate from "UI5Tooling" to "UI5CLI".

Split @ui5/server/internal/sslUtil into read-only certificate loading and explicit certificate generation. The root-password prompt now belongs to the CLI command.

BREAKING CHANGE: "ui5 serve --https" no longer generates or installs a self-signed certificate when none exists. Run "ui5 certificate generate" first, or pass --key and --cert.

JIRA: CPOUI5FOUNDATION-1347

See: https://github.com/UI5/cli/pull/1532#issuecomment-5369711431

### DIFF
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -27,6 +27,8 @@ Or update your global install via: `npm i --global @ui5/cli@next`
 
 - **@ui5/cli: `ui5 serve` renders a status banner in interactive terminals**
 
+- **@ui5/cli: New `ui5 certificate generate` command; `ui5 serve --https` no longer generates a certificate on the fly**
+
 - **@ui5/cli: Option `--experimental-css-variables` has been removed**
 
 - **@ui5/server: The internal Express server has been upgraded from Express 4 to Express 5**
@@ -383,9 +385,19 @@ The `--h2` option for the `ui5 serve` command has therefore been removed. Use th
 ui5 serve --https
 ```
 
-Like the previous `--h2` option, `--https` requires an SSL certificate and guides you through the automatic generation process on first use. See the [UI5 Server documentation](../pages/Server.md#ssl-certificates) for details.
+Like the previous `--h2` option, `--https` requires an SSL certificate. See [SSL Certificate Generation](#ssl-certificate-generation) below to learn how to generate a certificate or provide an existing one.
 
 If you need HTTP/2 for local development, manually set up a reverse proxy, for example, [nginx](https://nginx.org/), in front of the UI5 Server to handle the HTTP/2 connection with the browser.
+
+## SSL Certificate Generation
+
+In previous versions, UI5 CLI generated and installed a self-signed SSL certificate automatically. In UI5 CLI v5, certificate generation is an explicit step:
+
+```sh
+ui5 certificate generate
+```
+
+For certificate locations, custom paths, and server behavior when no certificate is found, see [SSL Certificates](../pages/Server#ssl-certificates).
 
 ## `sap-ui-version.json`
 
--- a/internal/documentation/docs/updates/migrate-v5.md
+++ b/internal/documentation/docs/updates/migrate-v5.md
@@ -391,7 +391,7 @@ If you need HTTP/2 for local development, manually set up a reverse proxy, for e
 
 ## SSL Certificate Generation
 
-In previous versions, UI5 CLI generated and installed a self-signed SSL certificate automatically. In UI5 CLI v5, certificate generation is an explicit step:
+In previous versions, UI5 CLI generated and installed a self-signed SSL certificate proactively. In UI5 CLI v5, certificate generation is an explicit step:
 
 ```sh
 ui5 certificate generate
--- a/packages/cli/lib/cli/commands/certificate.js
+++ b/packages/cli/lib/cli/commands/certificate.js
@@ -1,0 +1,97 @@
+import chalk from "chalk";
+import process from "node:process";
+import baseMiddleware from "../middlewares/base.js";
+import {getUi5DataDirOrDefault, getServerCertificatePaths, formatPath} from "../../dataDir.js";
+import {exists} from "../../utils/fsHelper.js";
+
+const certificateCommand = {
+	command: "certificate",
+	describe: "Manage the UI5 CLI server certificate",
+	middlewares: [baseMiddleware],
+};
+
+certificateCommand.builder = function(cli) {
+	return cli
+		.demandCommand(1, "Command required. Available command is 'generate'")
+		.command("generate", "Generate a self-signed server certificate and install it into the trust store", {
+			handler: handleGenerate,
+			builder: function(yargs) {
+				return yargs
+					.option("key", {
+						describe: "Path the private key is written to",
+						defaultDescription: "<UI5 data dir>/server/server.key",
+						type: "string"
+					})
+					.option("cert", {
+						describe: "Path the certificate is written to",
+						defaultDescription: "<UI5 data dir>/server/server.crt",
+						type: "string"
+					})
+					.option("force", {
+						alias: "f",
+						describe: "Generate a new certificate even if one already exists at the target path",
+						default: false,
+						type: "boolean"
+					})
+					.example("$0 certificate generate",
+						"Generate a server certificate in the default UI5 data directory")
+					.example("$0 certificate generate --force",
+						"Regenerate the server certificate, overwriting an existing one")
+					.example("UI5_DATA_DIR=/custom/path $0 certificate generate",
+						"Generate a server certificate in a non-default UI5 data directory");
+			},
+			middlewares: [baseMiddleware],
+		});
+};
+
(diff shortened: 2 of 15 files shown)


packages/cli/lib/cli/commands/serve.js
```diff
@@ -54,12 +53,12 @@ serve.builder = function(cli) {
 		})
 		.option("key", {
 			describe: "Path to the private key",
-			default: path.join(os.homedir(), ".ui5", "server", "server.key"),
+			defaultDescription: "<UI5 data dir>/server/server.key",
```
d3xter666: I was wondering about this `<UI5 data dir>`. To me it's clear, but I was wondering for people not that familiar wih the CLI.

Since, we mask the home dir with `~` now, wouldn't it be clearer if we just write it like that `~/.ui5/server/server.key`.

I would agree that `<UI5 data dir>` is rather an astract as it relies on `UI5_DATA_DIR`, a configuration and a default value, but if those are not provided, then `~/.ui5/server/server.key` is the location.

This is not a blocker from my side, just wanted to share my thoughts on it

packages/cli/lib/cli/commands/serve.js
```diff
@@ -231,10 +230,34 @@ serve.handler = async function(argv) {
...
-		const {getSslCertificate} = await import("@ui5/server/internal/sslUtil");
-		const {key, cert} = await getSslCertificate(serverConfig.key, serverConfig.cert);
-		serverConfig.key = key;
-		serverConfig.cert = cert;
+		// A default certificate path is only needed for HTTPS, so the UI5 data directory is
+		// resolved once here rather than for every serve invocation.
+		const ui5DataDir = await getUi5DataDirOrDefault({cwd: process.cwd()});
+		const defaults = getServerCertificatePaths(ui5DataDir);
+		const keyPath = serverConfig.key ?? defaults.keyPath;
+		const certPath = serverConfig.cert ?? defaults.certPath;
```
d3xter666: I see this pattern repeats and for me this seems like the natural way to resolve the `key` and `cert`. From that perspective, wouldn't it be easier if we get the resolved paths directly from `getServerCertificatePaths`

packages/cli/lib/cli/commands/serve.js
matz3: Thanks for the comment. I'm also not totally happy with this state, but it makes it clear that the default depends on the data dir, which is not always `~/.ui5`. That was the reason to not print `~/.ui5` anymore.

But I'm also fine with using `~/.ui5/server/`, although it then misses to describe that it depends on the data dir location.

packages/cli/test/lib/cli/commands/certificate.js
```diff
@@ -0,0 +1,238 @@
...
+		"../../../../lib/utils/fsHelper.js": {
+			exists: t.context.existsStub
+		}
+	});
+});
+
+test.afterEach.always((t) => {
+	sinon.restore();
+	esmock.purge(t.context.certificate);
+	delete process.env.UI5_DATA_DIR;
```
d3xter666: The wipe in the `beforeEach` block seems correct, but this one here not that much.
Currently, the `env` is "covered" by the fact that AVA tests run in an isolated scope under a different "thread". 
`env` is a global and only that isolation keeps us from poluting the global space.
I would assume that we stash and purge the `env.UI5_DATA_DIR` in `beforeEach`, but here we restore the stashed value instead of clearing it up.

packages/cli/lib/cli/commands/serve.js
matz3: Yeah, that makes sense. Done.

packages/cli/test/lib/cli/commands/certificate.js
matz3: Correct. I cleaned it up to align it with other existing tests which also restore it. I've also updated the `cache` test as it used the same pattern.

packages/cli/lib/cli/commands/serve.js
d3xter666: Yep, exactly, that's the struggle! I'm ok with the current solution, I just wanted to share my thoughts on that.

matz3: I would actually prefer your suggestion. We mention the UI5 data dir specifics in our documentation, and also in the command examples (generate with custom data dir).

In addition, the `<UI5 data dir>` caused issues with our CLI documentation script. The script should still be improved, and I want to look into that, but for now it would not be a blocker.

d3xter666 (dismissed):
LGTM

internal/documentation/docs/updates/migrate-v5.md
```diff
@@ -383,10 +385,20 @@ The `--h2` option for the `ui5 serve` command has therefore been removed. Use th
...
 ```
 
-Like the previous `--h2` option, `--https` requires an SSL certificate and guides you through the automatic generation process on first use. See the [UI5 Server documentation](../pages/Server.md#ssl-certificates) for details.
+Like the previous `--h2` option, `--https` requires an SSL certificate. See [SSL Certificate Generation](#ssl-certificate-generation) below to learn how to generate a certificate or provide an existing one.
 
 If you need HTTP/2 for local development, manually set up a reverse proxy, for example, [nginx](https://nginx.org/), in front of the UI5 Server to handle the HTTP/2 connection with the browser.
 
+## SSL Certificate Generation
+
+In previous versions, UI5 CLI generated and installed a self-signed SSL certificate automatically. In UI5 CLI v5, certificate generation is an explicit step:
```
RandomByte:
```suggestion
In previous versions, UI5 CLI generated and installed a self-signed SSL certificate proactively. In UI5 CLI v5, certificate generation is an explicit step:
```

packages/cli/lib/cli/commands/serve.js
```diff
@@ -231,10 +230,35 @@ serve.handler = async function(argv) {
...
+
+		const {getSslCertificate, SslCertificateNotFoundError} = await import("@ui5/server/internal/sslUtil");
+		try {
+			const {key, cert} = await getSslCertificate(keyPath, certPath);
+			serverConfig.key = key;
+			serverConfig.cert = cert;
+		} catch (err) {
+			if (err instanceof SslCertificateNotFoundError) {
+				const keyOrigin = serverConfig.key ? "--key" : "default";
+				const certOrigin = serverConfig.cert ? "--cert" : "default";
```
RandomByte:
```suggestion
				const keyOrigin = serverConfig.key ? "--key" : "default location";
				const certOrigin = serverConfig.cert ? "--cert" : "default location";
```

packages/cli/lib/cli/commands/serve.js
```diff
@@ -231,10 +230,35 @@ serve.handler = async function(argv) {
...
+				throw new Error(
+					`No SSL certificate found for HTTPS.\n` +
```
RandomByte:
```suggestion
					`Failed to find required SSL certificate for launching the HTTPS server.\n` +
```

I think our error messages more commonly start like this. I.e. "Failed to do x"

RandomByte (changes requested):
Just some wording feedback/proposals. See my earlier comment regarding Firefox for an actual issue.

RandomByte (approved):

packages/cli/lib/cli/commands/certificate.js
```diff
@@ -0,0 +1,104 @@
...
+		}
+	}
+
+	// Inform the user before triggering the trust-store installation, which requires elevated
+	// privileges and therefore prompts for the root password (or shows a confirmation dialog on Windows).
+	if (process.platform === "win32") {
+		process.stderr.write("Please press allow in the opened dialog to confirm importing the newly created " +
+			"SSL certificate into the operating system and browsers.\n");
+	} else {
+		process.stderr.write("Please enter your root password to allow importing the newly created " +
```
KlattG, line 84:
```suggestion
		process.stderr.write("Enter your root password to import the newly created " +
```

packages/cli/lib/cli/commands/certificate.js
```diff
@@ -0,0 +1,104 @@
...
+				`Use ${chalk.bold("--force")} to generate a new certificate and overwrite the existing one.\n`
+			);
+			return;
+		}
+	}
+
+	// Inform the user before triggering the trust-store installation, which requires elevated
+	// privileges and therefore prompts for the root password (or shows a confirmation dialog on Windows).
+	if (process.platform === "win32") {
+		process.stderr.write("Please press allow in the opened dialog to confirm importing the newly created " +
```
KlattG, line 81:
```suggestion
		process.stderr.write("Press 'Allow' in the opened dialog to confirm importing the newly created " +
```

KlattG (commented):
[LATE] We usually don't write 'Please' unless we're asking the user to do something unusually difficult or laborious

PR #1561, opened 2026-09-02
build(deps-dev): In-range update of dependencies

d3xter666:
Resolves: GHSA-73wf-gq98-2v4g


matz3 (approved):
AFAICS those are only devDependencies, so `build(deps-dev)` would be more suitable.

PR #1564, opened 2026-09-02
feat(cli): Add 'cache clean --project' to delete a project's build cache

RandomByte:
This can be helpful when debugging the incremental build. It allows to delete individual cache entries related to a project instead of the entire cache.

JIRA: CPOUI5FOUNDATION-1244

### DIFF
--- a/packages/cli/lib/cli/commands/cache.js
+++ b/packages/cli/lib/cli/commands/cache.js
@@ -51,7 +51,7 @@ cacheCommand.builder = function(cli) {
 						"Remove all cached UI5 data without confirmation (e.g. in CI scenarios)")
 					.example("$0 cache clean --project",
 						"Remove only the build cache of the project in the current directory")
-					.example("$0 cache clean --project sap.ui.core",
+					.example("$0 cache clean --project @openui5/sap.ui.core",
 						"Remove only the build cache of the project 'sap.ui.core'")
 					.example("UI5_DATA_DIR=/custom/path $0 cache clean",
 						"Remove cached data from a non-default UI5 data directory");
--- a/packages/cli/lib/cli/commands/cache.js
+++ b/packages/cli/lib/cli/commands/cache.js
@@ -3,12 +3,15 @@ import path from "node:path";
 import process from "node:process";
 import {isLogLevelEnabled} from "@ui5/logger";
 import baseMiddleware from "../middlewares/base.js";
+import {applyProjectConfigOptions, applyWorkspaceOptions} from "../options.js";
 import {getUi5DataDirOrDefault, formatPath} from "../../dataDir.js";
 import {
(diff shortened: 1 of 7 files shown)


packages/cli/lib/cli/commands/cache.js
```diff
@@ -32,10 +37,22 @@ cacheCommand.builder = function(cli) {
...
+							"project id (e.g. --project sap.ui.core) to target a specific project",
+						type: "string",
+					})
 					.example("$0 cache clean",
 						"Remove all cached UI5 data after confirmation")
 					.example("$0 cache clean --force",
 						"Remove all cached UI5 data without confirmation (e.g. in CI scenarios)")
+					.example("$0 cache clean --project",
+						"Remove only the build cache of the project in the current directory")
+					.example("$0 cache clean --project sap.ui.core",
```
d3xter666: This one doesn't seem to be correct. Project name seems to be the fully scoped name
```suggestion
					.example("$0 cache clean --project @openui5/sap.ui.core",
```

d3xter666 (commented):
I tried to break the solution in the following ways and it behaved correctly in all cases.

### 1. SQL injection via `--project`

Not possible — all queries that accept `projectId` use prepared statements with `?` placeholders. The value is bound, never interpolated into SQL.

### 2. Parallel cleanup scenarios

| Scenario | Result |
|---|---|
| Two project cleanups in parallel (e.g. `@openui5/sap.m` and `sap.ui.core`) | ✅ Works correctly by design — processes operate on disjoint rows |
| Full cleanup running in parallel with a project cleanup | ✅ Both end states are equivalent; full cleanup removes everything regardless of ordering |


### 3. Leaving stale records after an interrupted run

Not possible, for two reasons:

- All deletes are wrapped in a `BEGIN`/`COMMIT`/`ROLLBACK` transaction — a partial run is automatically rolled back.
- SQLite WAL mode provides crash recovery: if the process is terminated before `COMMIT` (e.g. via `SIGTERM`), the incomplete transaction has no commit record in the WAL and is discarded on the next open.

Reference: https://www.sqlite.org/wal.html

The per-project record set is also small enough that the delete completes in milliseconds, which further narrows the interruption window.

---

## Observations

### Silent no-op on unknown project name

Users need to know the **exact** project name (the `package.json` `name` field) to use `--project` effectively. For example, `sap.m` must be passed as `@openui5/sap.m`. If an incorrect or incomplete name is supplied:

```sh
ui5 cache clean --project my-project-nmae --force
```

There is no error and no output — indistinguishable from a successful cleanup. This is consistent with the existing design (no output without `--verbose`), but it creates a **silent no-op on a typo**.

packages/cli/lib/cli/commands/cache.js
RandomByte: Nice catch, thanks!

PR #1577, opened 2026-09-07
refactor(server): Unify HTTP/HTTPS server creation in listener

matz3:
Add a symmetric createServer helper to httpListener.js that builds the right Node server for the config, replacing the inline http.createServer / addSsl asymmetry in Supervisor.#init. addSsl folds into createServer, and listen's parameter is renamed app -> server to reflect that it binds a pre-created server rather than an Express app.

Express now stays confined to stack.js; the Supervisor no longer imports node:http directly.


codeworrior (approved):
Looks good and much cleaner to me.

PR #1578, opened 2026-09-07
fix(project): Memoize in-flight native watcher load to avoid polling race

matz3:
Concurrent subscribe() calls (as issued by WatchHandler via Promise.all) raced on the native-backend memoization: nativeBackendLoaded was flipped to true before the dynamic import() resolved, so callers settling before the winner assigned nativeBackend read a null value and silently fell back to the polling backend. On projects with many watched paths (e.g. the OpenUI5 testsuite), that meant nearly all source watchers polled ~42k files every 250ms, pegging idle CPU at ~250%.

Memoize the in-flight promise instead of a pre-await boolean so concurrent callers await the same import and all observe the resolved backend. Add a regression test that fires concurrent subscribe() calls before the first load resolves and asserts all use the native backend.


codeworrior (approved):
LGTM!

PR #1596, opened 2026-09-15
build(deps-dev): Replace adm-zip with node-stream-zip in e2e-tests

matz3:
adm-zip appears unmaintained; node-stream-zip is already used elsewhere in the repo, so this aligns on a single zip dependency.


codeworrior (approved):
LGTM.

